PR #1, opened 2016-08-31
Dev

dennisroberts71:
Add configuration defaults to terrain.

### DIFF
--- a/src/terrain/util/config.clj
+++ b/src/terrain/util/config.clj
@@ -24,26 +24,26 @@
   []
   (cc/mask-config props :filters [#"(?:irods)[-.](?:user|pass|key|secret)"]))
 
-(cc/defprop-int listen-port
+(cc/defprop-optint listen-port
   "The port that terrain listens to."
   [props config-valid configs]
-  "terrain.app.listen-port")
+  "terrain.app.listen-port" 60000)
 
-(cc/defprop-str environment-name
+(cc/defprop-optstr environment-name
   "The name of the environment that this instance of Terrain belongs to."
   [props config-valid configs]
-  "terrain.app.environment-name")
+  "terrain.app.environment-name" "docker-compose")
 
-(cc/defprop-vec allowed-groups
+(cc/defprop-optvec allowed-groups
   "The names of the groups that are permitted to access secured admin services."
   [props config-valid configs]
-  "terrain.cas.allowed-groups")
+  "terrain.cas.allowed-groups" ["core-services", "tito-admins", "tito-qa-admins", "dev"])
 
-(cc/defprop-str uid-domain
+(cc/defprop-optstr uid-domain
   "The domain name to append to the user identifier to get the fully qualified
    user identifier."
   [props config-valid configs]
-  "terrain.uid.domain")
+  "terrain.uid.domain" "cyverse.org")
 
 (cc/defprop-optboolean admin-routes-enabled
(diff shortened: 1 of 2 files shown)


src/terrain/util/config.clj
```diff
@@ -24,26 +24,26 @@
...
-  "terrain.cas.allowed-groups")
+  "terrain.cas.allowed-groups" ["core-services", "tito-admins", "tito-qa-admins", "dev"])
 
-(cc/defprop-str uid-domain
+(cc/defprop-optstr uid-domain
   "The domain name to append to the user identifier to get the fully qualified
    user identifier."
   [props config-valid configs]
-  "terrain.uid.domain")
+  "terrain.uid.domain" "cyverse.org")
```
ianmcorvidae: One nagging feeling I have is that some places we have `iplantcollaborative.org` hardcoded, which might make that the more sensible default -- not to mention, of course, that everything in our existing databases has that.

src/terrain/util/config.clj
```diff
@@ -336,141 +336,140 @@
...
   [props config-valid configs data-routes-enabled]
-  "terrain.infosquito.es-url")
+  "terrain.infosquito.es-url" "http://elasticsearch:9200")
 
-(cc/defprop-str jwt-private-signing-key
+(cc/defprop-optstr jwt-private-signing-key
   "The path to the private key used for signing JWT assertions."
   [props config-valid configs]
-  "terrain.jwt.signing-key.private")
+  "terrain.jwt.signing-key.private" "/path/to/private-key.pem")
```
ianmcorvidae: This doesn't seem like a useful default -- nobody's system is likely to have a `/path` directory.

src/terrain/util/config.clj
dennisroberts71: My thought was that we'd be switching to `cyverse.org` eventually anyway. We can override it for the time being.

ianmcorvidae:
Other than the couple of things I've commented on, and analogues in other parts of the config, looks good to me!

src/terrain/util/config.clj
ianmcorvidae: Yeah, I suppose that works. At least looking at this repository, I don't see any hardcoding, so I think my nagging feeling was amiss

src/terrain/util/config.clj
dennisroberts71: True. I was a little ambivalent about this change and others like it because on one hand, it would be nice to have the actual paths in these settings so that we could rely on the defaults, but was reluctant to supply this information to others. I suppose that wouldn't be _too_ risky, though. If someone actually gains access to the files, we have bigger problems anyway.

dennisroberts71:
Thanks for the review.

src/terrain/util/config.clj
ianmcorvidae: Yeah; I wasn't quite sure of that situation. I think that either it should be a real/recommended path of some sort, or it should be a required setting if we're not comfortable recommending anything, but I'm not sure which of those is more appropriate in which situations. I'm just not comfortable putting in a default that can't reasonably be used as an actual value of the config.

psarando:
Just a nitpick, but could this PR get a descriptive title? 😁

dennisroberts71:
Heh, that would be helpful. I added a brief description.

dennisroberts71:
Thanks for the reviews.

PR #2, opened 2016-09-06
CORE-6452 CORE-6453 Remove default-output-dir endpoints.

psarando:
These endpoints are no longer in use, and better logic for getting, setting, and creating the default output dir was implemented in the GET /secured/preferences, POST /secured/preferences, and GET /secured/bootstrap endpoints.


ianmcorvidae:
Straightforward, LGTM. :shipit:

psarando:
Thanks for the review 👍

PR #4, opened 2016-09-08
CORE-8062: removed the /apps/ids endpoint. which is not used

dennisroberts71:
This change removes an endpoint that is not used any longer.


ianmcorvidae:
straightforward, :shipit:

dennisroberts71:
Thanks for the review!

PR #5, opened 2016-09-15
CORE-8178 Update compojure-api version to 1.1.8

psarando:
Update versions for compojure-api and some of its dependencies.
Currently compojure-api is only used to wrap exception handling in terrain, so no routes had to be updated.


psarando:
Thanks for the review!

PR #8, opened 2016-11-09
CORE-7813 Reorganize /secured/bootstrap response

psarando:
The current `GET /secured/bootstrap` endpoint response has the following structure:
```json
{
    "loginTime": "1374190755304",
    "newWorkspace": false,
    "workspaceId": "4",
    "username": "snow-dog",
    "full_username": "snow-dog@iplantcollaborative.org",
    "email": "sd@example.org",
    "firstName": "Snow",
    "lastName": "Dog",
    "userHomePath": "/iplant/home/snow-dog",
    "userTrashPath": "/iplant/trash/snow-dog",
    "baseTrashPath": "/iplant/trash",
    "preferences": {
        "systemDefaultOutputDir": {
            "id": "/iplant/home/snow-dog/analyses",
            "path": "/iplant/home/snow-dog/analyses"
        },
        "defaultOutputFolder": {
            "id": "/iplant/home/snow-dog/analyses",
            "path": "/iplant/home/snow-dog/analyses"
        }
    },
    "auth-redirect": {
        "example": "https://example.org"
    }
}
```

These fields are pieced together from 3 different services other than terrain (apps, data-info, and preferences).

It would be helpful to group the info of each service under its own field, similar to the existing `preferences` field. Then if a service is down, rather than this terrain endpoint immediately throwing an error that prevents the UI from loading, the field that should contain that service's info will contain the error response instead. This will allow the UI to finish loading, with partial user info, and could also help the UI determine which service is down from the initial bootstrap endpoint response.

Since the response format is getting updated, the JSON keys can also be converted from camel-case to underscores (snake-case), which is more standard in our other endpoints.

For example:
```json
{
  "user_info": {
    "username": "snow-dog",
    "full_username": "snow-dog@iplantcollaborative.org",
    "email": "sd@example.org",
    "first_name": "Snow",
    "last_name": "Dog"
  },
  "session": {
    "login_time": 1374190755304,
    "auth_redirect": {
      "example": "https://example.org"
    }
  },
  "workspace": {
    "new_workspace": false,
    "workspace_id": "4"
  },
  "data_info": {
    "user_home_path": "/iplant/home/snow-dog",
    "user_trash_path": "/iplant/trash/snow-dog",
    "base_trash_path": "/iplant/trash"
  },
  "preferences": {
    "system_default_output_dir": {
      "id": "/iplant/home/snow-dog/analyses",
      "path": "/iplant/home/snow-dog/analyses"
    },
    "default_output_folder": {
      "id": "/iplant/home/snow-dog/analyses",
      "path": "/iplant/home/snow-dog/analyses"
    }
  }
}
```
Then if these services (besides terrain) are down, the bootstrap endpoint can still return success to the UI, but provide an error message letting the UI know which services are down or inaccessible.

For example:
```json
{
  "user_info": {
    "username": "snow-dog",
    "full_username": "snow-dog@iplantcollaborative.org",
    "email": "sd@example.org",
    "first_name": "Snow",
    "last_name": "Dog"
  },
  "session": {
    "error": "java.net.ConnectException: Connection refused"
  },
  "workspace": {
    "status": 400,
    "error": {
      "error_code": "ERR_ILLEGAL_ARGUMENT",
      "reason": {
        "foo": "disallowed-key"
      }
    }
  },
  "data_info": {
    "status": 404,
    "error": {
      "success": false,
      "reason": "unrecognized service path"
    }
  },
  "preferences": {
    "status": 503,
    "error": "<html>\r\n<head><title>503 Service Unavailable</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>503 Service Unavailable</h1></center>\r\n<hr><center>nginx/1.11.4</center>\r\n</body>\r\n</html>\r\n"
  }
}
```

### DIFF
--- a/src/terrain/services/bootstrap.clj
+++ b/src/terrain/services/bootstrap.clj
@@ -13,7 +13,7 @@
   [body]
   (try+
     (service/decode-json body)
-    (catch Throwable e
+    (catch Object _
       body)))
 
 (defn- trap-bootstrap-request
@@ -24,9 +24,9 @@
       (log/error e)
       {:status status
        :error  (decode-error-response body)})
-    (catch Throwable e
-      (log/error e)
-      {:error (str e)})))
+    (catch Object _
+      (log/error (:throwable &throw-context) "bootstrap request failed")
+      {:error (str (:throwable &throw-context))})))
 
 (defn- get-login-session
   [ip-address user-agent]
--- a/src/terrain/services/bootstrap.clj
+++ b/src/terrain/services/bootstrap.clj
@@ -1,0 +1,71 @@
+(ns terrain.services.bootstrap
+  (:use
+    [slingshot.slingshot :only [try+]]
+    [terrain.auth.user-attributes :only [current-user]])
+  (:require
+    [clojure.tools.logging :as log]
+    [terrain.clients.apps :as apps-client]
+    [terrain.clients.data-info :as data-info-client]
+    [terrain.services.user-prefs :as prefs]
+    [terrain.util.service :as service]))
+
+(defn- decode-error-response
+  [body]
+  (try+
+    (service/decode-json body)
+    (catch Throwable e
+      body)))
+
+(defn- trap-bootstrap-request
+  [req]
+  (try+
+    (req)
+    (catch #(not (nil? (:status %))) {:keys [status body] :as e}
+      (log/error e)
+      {:status status
+       :error  (decode-error-response body)})
+    (catch Throwable e
+      (log/error e)
+      {:error (str e)})))
+
+(defn- get-login-session
+  [ip-address user-agent]
+  (trap-bootstrap-request
+    #(select-keys (apps-client/record-login ip-address user-agent) [:login_time :auth_redirect])))
+
+(defn- get-workspace
+  []
+  (trap-bootstrap-request
+    #(select-keys (apps-client/get-workspace) [:id :new_workspace])))
+
+(defn- get-user-data-info
+  [user]
+  (trap-bootstrap-request
+    #(data-info-client/user-base-paths user)))
+
+(defn- get-user-prefs
+  [username]
+  (trap-bootstrap-request
+    #(prefs/user-prefs username)))
+
+(defn bootstrap
+  "This service obtains information about and initializes the workspace for the authenticated user.
+   It also records the fact that the user logged in."
+  [{{:keys [ip-address]} :params {user-agent "user-agent"} :headers}]
+  (service/assert-valid ip-address "Missing or empty query string parameter: ip-address")
+  (service/assert-valid user-agent "Missing or empty request parameter: user-agent")
+  (let [{user :shortUsername :keys [email firstName lastName username]} current-user
+        login-session (future (get-login-session ip-address user-agent))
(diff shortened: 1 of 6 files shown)


psarando:
Requires changes from cyverse-de/data-info#6 and cyverse-de/apps#17

ash3rz:
Is there something preventing us from having a consistent type of error JSON?  For `preferences` and `session` the error is just a string, but then `workspace` and `data_info` are objects that are different from each other.

psarando:
Regarding the different JSON errors from the apps and data-info services (`workspace` and `data_info` values in the example), I'll quote @johnworth 's response from chat:

> the formats are different because ... we (me, mostly) didn't think of standardizing until after the services were in place

As for the other two examples, those are just strings because terrain didn't get a JSON error response from the corresponding services.
* The `java.net.ConnectException: Connection refused` error message comes from an exception in terrain itself, because in this example it wasn't configured correctly, so it couldn't connect to the user-info service in the first place.
* The 503 error message in the `preferences.error` field is the HTML response nginx will return to terrain when the preferences service is down.

psarando:
Since the service error responses are not always a JSON  response, the UI has this [org.iplantc.de.commons.client.ErrorHandler#parseExceptionJson method](https://github.com/cyverse-de/ui/blob/2.7.0-RELEASE/de-lib/src/main/java/org/iplantc/de/commons/client/ErrorHandler.java#L154-L174) for displaying error messages in the error dialog details panel.

ianmcorvidae:
Just a couple comments in advance of any design discussioning we might do about this:

* The nginx error page responses can be modified. It should therefore not be an issue to make the 503 more consistent, we just have to put in in place. I've been wanting to do this anyway
* I'd love to standardize the JSON to some extent, but we should also have some amount of "add some arbitrary keys somewhere" ability -- not every detail we want to pass in an error is going to be formally defined, so an arbitrary object somewhere in the format would be good. Most stuff should be standardizable though, e.g. reasons etc.

psarando:
> I'd love to standardize the JSON to some extent, but we should also have some amount of "add some arbitrary keys somewhere" ability

I agree, our [clojure-commons.exception library](https://github.com/cyverse-de/iplant-clojure-commons/blob/2.8.1/src/clojure_commons/exception.clj) allows us to return a standard error JSON with at least `error_code` and `reason` fields, but allows arbitrary fields in addition to those. So it might be a good idea to start with this structure for all our service error responses?

src/terrain/services/bootstrap.clj
```diff
@@ -0,0 +1,71 @@
...
+
+(defn- trap-bootstrap-request
+  [req]
+  (try+
+    (req)
+    (catch #(not (nil? (:status %))) {:keys [status body] :as e}
+      (log/error e)
+      {:status status
+       :error  (decode-error-response body)})
+    (catch Throwable e
```
ianmcorvidae: It might be worth putting a `catch Object e` somewhere in here too. slingshot can throw basically anything, more than just Throwables, and it's possible we're using that somewhere.

psarando: Thanks, good catch.

psarando:
Thanks for the review 👍

PR #9, opened 2016-12-01
CORE-8462: changed the request body and the email format of the feedb…

dennisroberts71:
…ack endpoint

### DIFF
--- a/src/terrain/util/email.clj
+++ b/src/terrain/util/email.clj
@@ -81,30 +81,12 @@
       :template  "permanent_id_request_submitted"
       :values    template-values)))
 
-(defn- format-question
-  "Formats a question and answer for a user feedback submission."
-  [[q a]]
-  (let [q (string/replace q #"^(?![*])" "* ")
-        a (if (string? a) [a] a)]
-    (apply str "\n" q "\n" (mapv #(str % "\n") a))))
-
-(defn- feedback-email-text
-  "Formats the text for the user feedback email message."
-  [feedback]
-  (let [user (:username current-user)]
-    (apply str user " has provided some DE feedback:\n"
-           (mapv format-question feedback))))
-
 (defn send-feedback-email
   "Sends email messages containing user feedback."
-  [feedback]
-  (let [text (-> (assoc (dissoc feedback "shibbolethEppn")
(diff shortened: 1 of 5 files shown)


src/terrain/util/email.clj
```diff
@@ -81,30 +81,12 @@
...
-     :to        (config/feedback-dest-addr)
-     :from-addr (config/feedback-dest-addr)
-     :subject   "DE Feedback"
-     :template  "blank"
-     :values    {:contents text})))
+  [{:strs [email fields subject]}]
+  (send-email
+   :to        (config/feedback-dest-addr)
+   :from-addr (or email (:email current-user))
+   :subject   (or subject "DE Support Request")
```
psarando: The default subject for this `send-feedback-email` function used to be `DE Feedback`. Why change it to `DE Support Request`?

src/terrain/util/email.clj
```diff
@@ -81,30 +81,12 @@
...
+   :template  "blank"
+   :values    {:contents (apply str (mapv (fn [[k v]] (str k ": " v "\n")) fields))}))
```
psarando: My initial instinct would be to use `clojure.string/join` here.

Maybe something like
```clojure
(string/join "\n" (mapv (partial string/join ": ") fields))
```

Of course, this current method works just as well :)

dennisroberts71: Heh, that would work. I wanted to get the trailing newline in there, though. :)

psarando: Ah, good point, I missed that 👍

src/terrain/util/email.clj
dennisroberts71: This is part of the move away from having us receive "feedback" any time a user encounters a job failure. I suppose I could change the URL path to reflect something more accurate...

psarando: Ah OK, I knew we were adding user support emails, but I didn't realize we were getting rid of general feedback too.

psarando:
Looks good. I'd say :shipit:, but I don't know if we want to wait for UI changes to handle the new endpoint?

dennisroberts71:
Thanks for the review!

PR #10, opened 2016-12-01
CORE-8446 Fix /bootstrap error encoding.

psarando:
This PR includes fixes in the `/bootstrap` endpoint for parsing error responses from the apps and preferences services when they are under maintenance (503 or 504 responses).


psarando:
Thanks for the review 👍

PR #11, opened 2016-12-02
CORE-8443 Provide passthrough to apps for redirect uris

ash3rz:
The UI typically is informed of any redirect uris via the bootstrap endpoint.  Exposing this will allow the UI an opportunity to retrieve those uris outside of bootstrap.


ash3rz:
heh, yeah, I figured since Terrain is so foreign to me it's worth a PR 😄 
Thanks for the review!

PR #12, opened 2016-12-06
CORE-8462: allow support-email endpoint to accept fields with multipl…

dennisroberts71:
…e values


dennisroberts71:
Thanks for the review!

PR #19, opened 2017-01-19
For 503s when calling data-info in paged-directory, rethrow as another 503

ianmcorvidae:
This should make paged-directory properly mark data-info as down with the new UI resiliency stuff.


ianmcorvidae:
(this is quite simple so I'll merge right away, but I thought I'd open it for later reference if desired/if people wanted to review anyway)

PR #22, opened 2017-04-29
CORE-8720 Add new apps endpoints for private tools.

psarando:
The following apps service endpoint passthroughs have been added:

* `GET /admin/tools`
* `GET /admin/tools/:tool-id`
* `POST /tools`
* `POST /tools/permission-lister`
* `POST /tools/sharing`
* `POST /tools/unsharing`
* `DELETE /tools/:tool-id`
* `PATCH /tools/:tool-id`

Also the `public` parameter is now allowed with the `GET /tools` endpoint.

### DIFF
--- a/src/terrain/clients/apps.clj
+++ b/src/terrain/clients/apps.clj
@@ -27,9 +27,9 @@
       (:body)
       (service/decode-json)))
 
-(defn import-tools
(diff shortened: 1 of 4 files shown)


psarando:
Thanks for the review 👍

PR #23, opened 2017-05-06
Added services to add and list collaborator lists.

dennisroberts71:


src/terrain/clients/iplant_groups.clj
```diff
@@ -57,3 +60,49 @@
...
+
+(defn- get-or-add-folder [client user name]
+  (or (get-optional-folder client user name)
+      (create-user-folder client user name)))
+
+(defn- get-collaborator-list-folder [user client]
+  (let [name (c/build-folder-name client (format "users:%s:collaborator-lists" user))]
+    (get-or-add-folder client user (string/replace name  #":[^:]+$" ""))
+    (get-or-add-folder client user name)))
+
```
dennisroberts71: Your understanding of this is correct.

I tend to use `or` for things like this quite a bit especially when external API calls are involved, but your changes make sense to me as well. I'll make these changes.

src/terrain/clients/iplant_groups.clj
```diff
@@ -57,3 +60,49 @@
...
+;; update to iplant-groups will be required to eliminate this hack.
+(defn get-collaborator-lists
+  ([user]
+   (let [client (get-client)
+         folder (:name (get-collaborator-list-folder user client))]
+     {:groups (mapv (partial format-collaborator-list folder) (:groups (c/find-groups client user folder)))}))
+  ([user search]
+   (let [client (get-client)
+         folder (:name (get-collaborator-list-folder user client))]
+     {:groups (mapv format-collaborator-list (c/find-groups client user search folder))})))
```
dennisroberts71: Good catch. I'll fix that.

src/terrain/clients/iplant_groups.clj
psarando: I think we also still want to create the `collab-folder` if it doesn't exist, so the current `get-or-add-folder` covers both the user folder and the collab folder.

I'm used to the `or` short-circuiting expression, but I can see how it could be trickier to read, and when I first reviewed this PR, I also almost suggested this change:

```
(defn- get-or-add-folder [client user name]
  (try+
    (c/get-folder client user name)
    (catch [:status 404] _
      (create-user-folder client user name))))
```

dennisroberts71: @psarando is correct; we need to ensure that the collaborator list folder exists in either case.

src/terrain/clients/iplant_groups.clj
```diff
@@ -64,45 +64,53 @@
...
-    (c/grant-folder-privilege client (config/grouper-user) name user :stem)
-    folder))
+(defn- get-user-folder-name [client user]
+  (c/build-folder-name client (format "users:%s" user)))
 
-(defn- get-or-add-folder [client user name]
-  (or (get-optional-folder client user name)
-      (create-user-folder client user name)))
+(defn- get-collaborator-list-folder-name [client user]
+  (c/build-folder-name client (format "users:%s:collaborator-lists" user)))
```
psarando: I like these separate functions for building user 🆚 folder names 👍

dennisroberts71:
Thanks for the review!

PR #24, opened 2017-05-10
CORE-8800: get and delte endpoints for specific collaborator lists.

dennisroberts71:
This change adds `GET /collaborator-lists/:name` and `DELETE /collaborator-lists/:name` endpoints. It also changes the endpoints so that they don't try to create the Grouper folder that is used for collaborator lists unless that folder actually has to be present. For the time being, the only endpoint that has to ensure that the folder exists is the endpoint that adds a new collaborator list. I believe that this is the only endpoint that will have to ensure that the folder exists, but I'm not willing to commit to that statement completely at this time.

I'm hoping that preventing multiple endpoints from trying to create the folder will eliminate a concurrency issue that we were encountering when the listing and addition endpoints were called at about the same time. We may still encounter the same concurrency issue when there are two separate and nearly simultaneous calls to add a collaborator list for the same user when the collaborator-list folder hasn't been created yet. This situation should be rare enough not to cause significant problems for the time being. We may have to revisit this when the public API is exposed, though.

### DIFF
--- a/src/terrain/clients/iplant_groups.clj
+++ b/src/terrain/clients/iplant_groups.clj
@@ -69,11 +69,16 @@
   (c/grant-folder-privilege client (config/grouper-user) name user :stem)
   nil)
 
-(defn- ensure-folder-exists [client user name]
+(defn- folder-exists? [client user name]
   (try+
    (c/get-folder client user name)
+   true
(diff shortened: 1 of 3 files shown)


dennisroberts71:
Thanks for the review! 👍

PR #25, opened 2017-05-11
CORE-8830: added a subject search endpoint

dennisroberts71:


src/terrain/clients/iplant_groups.clj
```diff
@@ -49,21 +49,14 @@
     user-info
     (empty-user-info short-username)))
 
-(defn search-subjects
-  "Uses iplant-groups's subject search endpoint to retrieve user details."
-  [user search]
-  (let [res (http/get (str (curl/url (config/ipg-base) "subjects"))
-                      ;; Adding wildcards matches previous (trellis) search behavior
-                      {:query-params {:user user :search (str "*" search "*")}
```
dennisroberts71: I couldn't produce different results with or without wildcards, so the replacement for this function simply leaves them off.

ianmcorvidae: This is odd -- I recall that when I was doing this before, some stuff changed and some stuff didn't. It may be that the search methods in the grouper config have since changed, though, since I recall we're also able to control whether wildcards are added there or not.

dennisroberts71: Yeah, we've moved to a slightly newer version of Grouper since then, but that's the only change that I can think of that would have caused this.

ianmcorvidae: (you're right though. I'm just wondering what changed)

dennisroberts71:
Thanks for the review! 👍

PR #27, opened 2017-05-24
CORE-8864: added a new PATCH /collaborator-lists/:name endpoint

dennisroberts71:


src/terrain/routes/collaborator.clj
```diff
@@ -23,6 +23,9 @@
    (GET "/collaborator-lists/:name" [name]
      (service/success-response (cl/get-collaborator-list current-user name)))
 
+   (PATCH "/collaborator-lists/:name" [name :as {:keys [body]}]
+     (service/success-response (cl/update-collaborator-list current-user name (json/decode (slurp body) true))))
```
psarando: We could use `(terrain.util.service/decode-json body)` here?

dennisroberts71: Ah! I'd completely forgotten about that! I'll update it, thanks.

dennisroberts71:
Thanks for the review! 👍

PR #28, opened 2017-05-31
CORE-8866: modified the GET /secured/user-info and GET /secured/user-…

dennisroberts71:
…search endpoints so that they return user information in the same format as iplant-groups


dennisroberts71:
Thanks for the review! 👍

PR #29, opened 2017-06-02
CORE-8874: ensure that the permissions service is updated when a coll…

dennisroberts71:
…aborator list is deleted


dennisroberts71:
Thanks for the review!

PR #31, opened 2017-06-16
Delete and update endpoints for teams.

dennisroberts71:


src/terrain/clients/iplant_groups.clj
```diff
@@ -79,7 +79,11 @@
     (create-folder client user name))
   nil)
 
-;;
+;; General group functions.
+
+(defn- full-group-name [name folder]
+  (when name
+    (format "%s:%s" folder name)))
```
dennisroberts71: The argument order kinda seems backwards here, but I wanted to be able to easily use this function with `update`.

dennisroberts71:
Thanks for the review!

PR #32, opened 2017-06-19
CORE-8913: create collaborator lists automatically when a user adds a…

dennisroberts71:
… member to a list that does not exist


dennisroberts71:
This is a small change, so I'm going to merge it now. Reviews are still welcome, though.

PR #33, opened 2017-06-20
Team Membership Endpoints

dennisroberts71:
This change adds endpoints to add, remove, and list team members. It also adds an extra validation step to several existing endpoints to ensure that a 404 is returned if someone tries to update a team that doesn't exist.


dennisroberts71:
Thanks for the review! 👍

PR #35, opened 2017-06-27
Automatic addition and removal of `optout` privilege for team members.

dennisroberts71:


project.clj
```diff
@@ -34,7 +34,7 @@
                  [org.cyverse/clj-icat-direct "2.8.0"]
                  [org.cyverse/clj-jargon "2.8.3"]
                  [org.cyverse/clojure-commons "2.8.1"]
-                 [org.cyverse/cyverse-groups-client "0.1.0-SNAPSHOT"]
+                 [org.cyverse/cyverse-groups-client "0.1.2-SNAPSHOT"]
```
dennisroberts71: Heh, I created a couple of premature releases here.

src/terrain/clients/iplant_groups.clj
```diff
@@ -224,13 +224,18 @@
...
+                                        {:subject_id c/public-user :privileges public-privileges}]}))
+
+(defn add-team [user {:keys [name description public_privileges] :or {public_privileges []}}]
   (let [client (get-client)
         folder (get-team-folder-name client user)]
     (ensure-team-folder-exists client user)
     (let [full-name (str folder ":" name)
           group     (c/add-group client user full-name team-group-type description)]
-      (dorun (map (partial c/grant-group-privilege client user full-name c/public-user) public_privileges))
+      (grant-initial-team-privileges client user full-name public_privileges)
```
dennisroberts71: Since the bulk permission update endpoint is available now, I figured we might as well use it here. This change also grants admin privileges to the `de_grouper` user, which is necessary to add and revoke the `optout` privilege when users join or leave a group.

src/terrain/clients/iplant_groups.clj
```diff
@@ -266,20 +271,49 @@
...
     (c/remove-group-members client user group members)))
 
+(defn join-team [user name]
+  (let [client (get-client)
+        folder (get-team-folder-name client)
+        group  (full-group-name name folder)]
+    (verify-group-exists client user group)
+    (let [response (c/add-group-members client user group [user])]
+      (grant-optout-privileges client (config/grouper-user) group [user])
+      response)))
```
dennisroberts71: The update order here allows Grouper to verify that the user has permission to join the group before granting the `optout` privilege to the user.

src/terrain/clients/iplant_groups.clj
```diff
@@ -266,20 +271,49 @@
...
+
+(defn leave-team [user name]
+  (let [client (get-client)
+        folder (get-team-folder-name client)
+        group  (full-group-name name folder)]
+    (verify-group-exists client user group)
+    (let [response (c/remove-group-members client user group [user])]
+      (revoke-optout-privileges client (config/grouper-user) group [user])
+      response)))
```
dennisroberts71: The update order here serves a similar purpose to the update order for `join-team` it's not as important here, however, because a permission is being removed.

dennisroberts71:
Thanks for the review!

PR #37, opened 2017-07-17
Consul configs

dennisroberts71:


dennisroberts71:
I tested it using the `--dry-run` command-line option for `run-service`, which is very helpful. The resulting configuration setting looks correct to me. There was one missing setting (the default resource for iRODS) I assumed that that was because the setting is blank in Consul. The path to the configuration setting appears to be right to me. At any rate, the default setting for the default resource is the empty string, so this will work fine. I didn't actually try starting up the service locally, but doing the setup to get that to work without being able to connect to the overlay network takes more time than it's worth.

dennisroberts71:
Thanks for the review!

PR #38, opened 2017-07-17
CORE-8949 Prevent potential "Too many open files" error.

psarando:
Prevents potential `JargonException` "Unknown iRODS exception code recieved:-510024" in `GET /secured/favorites/filesystem` endpoint.


psarando:
Thanks for the review!

PR #40, opened 2017-08-01
CORE-8969: added the display_name field to group and collaborator lis…

dennisroberts71:
…t membership listings


dennisroberts71:
Thanks for the review!

PR #43, opened 2017-08-22
CORE-8984: team join request denial endpoint

dennisroberts71:


src/terrain/clients/notifications.clj
```diff
@@ -47,17 +47,28 @@
...
+  (send-notification {:type "team"
+                      :user (:id admin)
+                      :subject (str user-name " has requested to join team \"" team-name "\"")
+                      :email true
+                      :email_template "team_join_request"
+                      :payload {:email_address (:email admin)
+                                :requester_name user-name
+                                :requester_email user-email
+                                :requester_message message
+                                :team_name team-name}}))
```
dennisroberts71: Heh, I realized when I was looking over the code that I had a `try` without a `catch`. There's no real reason to allow this endpoint to succeed if the notification fails, so I just removed the `try`.

src/terrain/services/teams.clj
```diff
@@ -36,9 +37,15 @@
 (defn join [{user :shortUsername} name]
   (ipg/join-team user name))
 
-(defn join-request [{user :shortUsername user-name :commonName email :email :as user-info} name message]
+(defn join-request [{user :shortUsername user-name :commonName email :email} name message]
```
dennisroberts71: The `:as user-info` bit was actually debugging code because I was logging the user information. This variable is no longer used, so I removed it.

src/terrain/services/teams.clj
```diff
@@ -36,9 +37,15 @@
...
   (let [admin (first (ipg/get-team-admins user name))]
     (cn/send-team-join-notification user-name email name admin message)))
 
+(defn deny-join-request [{user :shortUsername} name requester message]
+  (ipg/verify-team-exists user name)
```
ash3rz: (Sorry if this is obvious...) So if the user hits this endpoint and the team doesn't exist, will the UI actually get a 404 back?  Is there an error message that lets the user know the team doesn't exist?

dennisroberts71: Ah, good question. The error propagates up from `iplant-groups`, so the response body indicates that the group doesn't exist:

```

ash3rz: 👍  cool, thanks for the explanation!

dennisroberts71:
thanks for the review!

PR #44, opened 2017-08-23
CORE-8985: only list directly assigned privileges for teams

dennisroberts71:


dennisroberts71:
This change is small, so I'm going to merge it immediately. Reviews are still welcome.

PR #45, opened 2017-08-24
CORE-8986: allow administrative users to remove themselves from a team

dennisroberts71:
This change only applies if a team administrator tries to remove themselves using the POST /teams/:team-name/members/deleter endpoint. The POST /teams/:team-name/leave endpoint will still work because it uses the configured Grouper admin user to update the team privileges.


dennisroberts71:
This change is small, so I'm going to merge it immediately. Reviews are still welcome.

PR #46, opened 2017-08-24
Team notification updates.

dennisroberts71:
The primary purpose of this PR is to notify users when they have been added to a team. A minor modification to the team join request notification was also required.

### DIFF
--- a/src/terrain/clients/notifications.clj
+++ b/src/terrain/clients/notifications.clj
@@ -76,7 +76,7 @@
   [user team-name]
   (send-notification {:type "team"
                       :user (:id user)
-                      :subject (str "Added to team")
+                      :subject "Added to team"
                       :email true
                       :email_template "added_to_team"
                       :payload {:email_address (:email user)
--- a/src/terrain/clients/notifications.clj
+++ b/src/terrain/clients/notifications.clj
@@ -70,6 +70,17 @@
                                 :team_name team-name
                                 :admin_message message}}))
 
+(defn send-team-add-notification
+  "Sends a notification indicating that the userhas been added to a team."
+  [user team-name]
+  (send-notification {:type "team"
+                      :user (:id user)
+                      :subject (str "Added to team")
+                      :email true
+                      :email_template "added_to_team"
+                      :payload {:email_address (:email user)
+                                :team_name team-name}}))
+
 (defn mark-all-notifications-seen
   []
   (raw/mark-all-notifications-seen (cheshire/encode (add-current-user-to-map {}))))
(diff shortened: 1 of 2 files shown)


src/terrain/services/teams.clj
```diff
@@ -23,7 +23,10 @@
   (ipg/get-team-members user name))
 
 (defn add-team-members [{user :shortUsername} name {:keys [members]}]
-  (ipg/add-team-members user name members))
+  (let [response (ipg/add-team-members user name members)]
```
ianmcorvidae: This is vanishingly minor and iplant-groups may not support it, but I'd imagine it's possible for not every member to successfully be added (e.g., if someone's already a member). Would it be possible to inspect the response and only send notifications for those members who were actually newly added?

src/terrain/clients/notifications.clj
```diff
@@ -70,6 +71,17 @@
                                 :team_name team-name
                                 :admin_message message}}))
 
+(defn send-team-add-notification
+  "Sends a notification indicating that the userhas been added to a team."
+  [user team-name]
+  (send-notification {:type "team"
+                      :user (:id user)
+                      :subject (str "Added to team")
```
ianmcorvidae: Probably don't need the `(str )` wrapper here, seems a bit confusing to keep it.

src/terrain/clients/notifications.clj
```diff
@@ -46,13 +46,14 @@
 
 (defn send-team-join-notification
   "Sends a notification indicating that a user has requested to join a team."
-  [user-name user-email team-name admin message]
+  [user user-name user-email team-name admin message]
```
ianmcorvidae: That `user` is a username and `user-name` is a user's full name is slightly confusing, but I don't know if there's a better way to do it. Perhaps something like `user-id` instead of `user`? But `user-name` is always going to be confusing whether it's a username (which we use as an ID) or a user's name, I suppose.

dennisroberts71: Heh, yeah, I briefly thought about that too.

src/terrain/clients/notifications.clj
dennisroberts71: Heh, oops! I was originally going to assemble a string and decided not to. I'll fix that.

src/terrain/services/teams.clj
dennisroberts71: `iplant-groups` currently returns an error if any member addition fails. I tested this, and the one correct member actually wasn't added to the team, which was rather surprising:

ianmcorvidae: Ah, interesting, so if any fail it backs out the whole set of changes. I suppose that's sensible of it.

How strange to have something upstream doing something that seems reasonable, heh

dennisroberts71:
Thanks for the review!

PR #47, opened 2017-08-25
CORE-8988: automatically grant read privileges when team members are …

dennisroberts71:
…added


dennisroberts71:
This change is tiny, so I'm going to merge it immediately. Reviews are still welcome.

PR #49, opened 2017-09-06
CORE-8995: also delete the permissions service resource when a team i…

slr71:
…s deleted


slr71:
Thanks for the reviews!

PR #51, opened 2017-10-25
CORE-9052: added the ability to search for app categories by name

slr71:
This is a new endpoint to search for app categories by exact name.


slr71:
Thanks for the review!

PR #52, opened 2017-10-27
CORE-9054: added the ability to permanently delete a metadata template

slr71:


slr71:
Thanks for the review!

PR #55, opened 2017-10-31
de-1129 Add route to webhook types endpoint

sri2k1us:


src/terrain/routes/webhooks.clj
```diff
@@ -10,4 +10,6 @@
     [config/app-routes-enabled]
 
     (PUT "/webhooks" [:as {:keys [body]}]
-      (service/success-response (apps-client/save-webhooks body)))))
+      (service/success-response (apps-client/save-webhooks body)))
+    (GET "/webhooks/types" [:as {:keys [params]}]
```
slr71: This isn't required, but I normally add a blank line between route definitions.

ianmcorvidae: I suspect you don't need anything inside the `[]` here, since you're not using it below.

slr71: Ah, good catch.

sri2k1us:
Thanks for the reviews. Pushed changes to address the feedback.

PR #57, opened 2017-11-07
CORE-9067: metadata cleanup endpoints.

slr71:
These new endpoints are intended primarily for use in the QA cleanup suite.


slr71:
Thanks for the review!

PR #59, opened 2017-11-16
Add a set of search service passthrough endpoints

ianmcorvidae:
(for data search only for now, of course)

This should be pretty straightforward? But, I want to make sure I'm still using best-practices since I haven't been doing much clojure lately.

### DIFF
--- a/src/terrain/clients/search.clj
+++ b/src/terrain/clients/search.clj
@@ -1,0 +1,18 @@
+(ns terrain.clients.search
+  (:use [terrain.util.transformers :only [secured-params]]
+        [terrain.util.config :only [search-base]])
+  (:require [clj-http.client :as http]
+            [cemerick.url :refer [url]]
+            [cheshire.core :as json]))
+
+(defn do-data-search
+  [params body]
+  (let [req-options  {:body         (json/encode body)
+                      :query-params params
+                      :as           :stream
+                      :content-type "application/json"}]
+    (http/post (str (url (search-base) "data" "search")) req-options)))
+
+(defn get-data-search-documentation
+  []
+  (http/get (str (url (search-base) "data" "documentation"))))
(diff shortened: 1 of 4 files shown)


ianmcorvidae:
Oh, and I haven't documented this yet, but that goes in other repos anyway.

slr71:
Looks good to me. :shipit:

ianmcorvidae:
Cool, thanks for the review!

src/terrain/clients/search.clj
```diff
@@ -0,0 +1,18 @@
+(ns terrain.clients.search
+  (:use [terrain.util.transformers :only [secured-params]]
```
psarando: It doesn't look like `secured-params` is used here yet (but we can leave it for now if we're planning on passing the user params to the search service in the future).

src/terrain/util/config.clj
```diff
@@ -504,6 +504,18 @@
...
+(cc/defprop-optstr search-base-url
+  "The hostname of the search service"
+  [props config-valid configs]
+  "terrain.search.base-url" "http://search:60000")
+
+(def search-base
+  (memoize
+   (fn []
+     (if (System/getenv "SEARCH_PORT")
+       (cfg/env-setting "SEARCH_PORT")
```
psarando: Should we name this `SEARCH_URL` or something? The `else` is returning `search-base-url`...

src/terrain/clients/search.clj
ianmcorvidae: oh, you're right. I originally used it, but the structuring using `terrain.util/controller` means I didn't need to, but I forgot to remove the import.

src/terrain/util/config.clj
ianmcorvidae: hm, I think we don't use the environment-based configs anyway; I was copying the others which have the same weirdness. I think I'll just remove this and use `search-base-url` directly

PR #60, opened 2017-11-21
CORE-9073: added the full-listing query parameter to the permission-l…

slr71:
…ister endpoints


slr71:
I added a new endpoint to allow administrators to remove users from the `de-users` group. This change is tracked under another Jira issue, but it's somewhat related because it's needed for the QA regression run.

slr71:
I just pushed a change to create another new endpoint. This one allows administrators to list groups that a user belongs to, and it's also needed for the QA regression run.

slr71:
Thanks for the review!

PR #61, opened 2018-03-14
CORE-9169 Email data admins when DOI data folder could not be moved.

psarando:
This PR will update the DOI request and creation functions so that errors from the `data-info` service will now be caught, data admins will be emailed about the error, and the DOI request/creation process can continue successfully.

This PR depends on cyverse-de/iplant-email#5

### DIFF
--- a/src/terrain/services/permanent_id_requests.clj
+++ b/src/terrain/services/permanent_id_requests.clj
@@ -166,21 +166,34 @@ For example, https://doi.org/10.7946/P2G596 links to the DOI 10.7946/P2G596.")
       (data-info-client/create-dirs (config/irods-user) [staging-path])
       (data-info/share (config/irods-user) [curators] [staging-path] "own"))))
 
+(defn- move-folder
+  [user {:keys [id path]} dest-path]
+  (try+
+   (data-info-client/move-single user id dest-path)
+   (ft/path-join dest-path (ft/basename path))
+
+   (catch Object e
+     (log/error e)
+     (email/send-permanent-id-request-data-move-error path dest-path current-user (:body e (str e)))
+     path)))
+
 (defn- stage-data-item
-  [user {:keys [id path]}]
-  (let [staged-path (format-staging-path path)
-        curators    (config/permanent-id-curators-group)]
-    (data-info-client/move-single (config/irods-user) id (config/permanent-id-staging-dir))
+  [user folder]
+  (let [curators-group (config/permanent-id-curators-group)
+        staged-path    (move-folder (config/irods-user)
+                                    folder
+                                    (config/permanent-id-staging-dir))]
+    (data-info/share (config/irods-user) [curators-group] [staged-path] "own")
     (data-info/share (config/irods-user) [user] [staged-path] "write")
-    (data-info/share (config/irods-user) [curators] [staged-path] "own")
     staged-path))
 
 (defn- publish-data-item
-  [user {:keys [id path]}]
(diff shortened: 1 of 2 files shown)


slr71:
I didn't catch any issues with this. :shipit:

slr71:
Just to make sure I understand this, it appears that in the case where the files are being placed in the staging directory, we revoke own permissions from the user on the files regardless of whether the files were moved successfully or not. If the files were not moved successfully, a member of the data curators group eventually moves the files using a DE job.

slr71:
The one question that I had was that typical DE jobs are incapable of moving files and instead must copy the files. I assume there's a special job available to the data curators group that correctly handles moving the files.

In the case where the files are being placed in the publishing directory, it appears to me that the use case is the nearly identical. The only major difference is that the user submitting the initial request is a member of the data curators group already.

psarando:
> Just to make sure I understand this, it appears that in the case where the files are being placed in the staging directory, we revoke own permissions from the user on the files regardless of whether the files were moved successfully or not.

Basically, except if the move was not successful, attempting to set `write` permissions on the folder for the original owner becomes a no-op, since `data-info` will see that user already has `own` permissions and will do nothing. If the move is successful, then the moved folder will be owned by the proxy user, so the permissions can then be reset to give the original owner `write` permissions and the data curators `own`.

> If the files were not moved successfully, a member of the data curators group eventually moves the files using a DE job.

The case that prompted this issue was when a user attempts to get a DOI for a folder with more than 1000 sub-files/folders. In that case the data curator would need to use icommands to move the folder since `data-info` will not allow it.

> The one question that I had was that typical DE jobs are incapable of moving files and instead must copy the files. I assume there's a special job available to the data curators group that correctly handles moving the files.

This endpoint will not use a DE job to move the DOI folder. It is using `data-info` as the iRODS proxy user to move the folder.

> In the case where the files are being placed in the publishing directory, it appears to me that the use case is the nearly identical. The only major difference is that the user submitting the initial request is a member of the data curators group already.

No, the original owner of the data who is requesting the DOI is supposed to lose ownership of the data, since it will be published in the public commons repo folder. The original owner still retains `write` permissions during the process in order to update metadata if required, but once the data is published, only the data curators can modify the public data.

Here is an overview of the DOI request process: https://cyverse-de.github.io/api/endpoints/permanent-id-requests.html#workflow-overview

slr71:
> Basically, except if the move was not successful, attempting to set write permissions on the folder for the original owner becomes a no-op, since data-info will see that user already has own permissions and will do nothing. If the move is successful, then the moved folder will be owned by the proxy user, so the permissions can then be reset to give the original owner write permissions and the data curators own.

Ah, that's a good point. It's been a while since I've looked at the sharing stuff.

> This endpoint will not use a DE job to move the DOI folder. It is using data-info as the iRODS proxy user to move the folder.

Ah, I read "as a job" in the Jira issue and assumed it was being done as a DE job.

> No, the original owner of the data who is requesting the DOI is supposed to lose ownership of the data...

Heh, I noticed that difference and glossed over it in my comment. 😊

slr71:
Thanks for the clarifications.

psarando:
> Ah, I read "as a job" in the Jira issue and assumed it was being done as a DE job.

Oh, good point. That issue should be part of a larger Epic for async `data-info` operations, and this PR will not address that.

I'll create a new issue for this PR, which is really a workaround for the bug that prompted CORE-9108 in the first place.

psarando:
Thanks for the review 👍

PR #63, opened 2018-06-13
CORE-9096: removed the `/callbacks/notification` endpoint.

slr71:


slr71:
Thanks for the review!

project.clj
```diff
@@ -1,6 +1,11 @@
 (use '[clojure.java.shell :only (sh)])
 (require '[clojure.string :as string])
 
+;; This should be removed as soon as the need for Forester is removed.
```
psarando: I didn't see anything related to `Forester` added in this PR. Is this still required?

slr71: Yeah, unfortunately it is. We use `Forester` for the `tree-urls` endpoints. The only reason I had to add that blurb in `project.clj` is because the version of Leiningen that I'm using refuses to use a Maven repository that doesn't use HTTPS unless it's there.

slr71: I omitted the `;; never do this!` comment from `lein help faq`. 😆

psarando: OK, thanks. I just wasn't sure if that was related to removing `/callbacks/notification` somehow 😀

PR #64, opened 2018-07-03
CORE-9249: added an OAuth authentication endpoint.

slr71:


src/terrain/util/config.clj
```diff
@@ -493,6 +493,21 @@
...
+  "The base URI for the OAuth API endpoints."
+  [props config-valid configs]
+  "terrain.oauth.base-uri" "https://cas/oauth2.0")
+
+(cc/defprop-optstr oauth-client-id
+  "The client ID to use when authenticating to the OAuth server."
+  [props config-valid configs]
+  "terrain.oauth.client-id" "fake-client-id")
+
+(cc/defprop-optstr oauth-client-secret
```
slr71: This configuration setting isn't used yet, but we'll need it when we add support for OAuth tokens to authenticated endpoints.

slr71:
Thanks for the review!

PR #67, opened 2018-07-26
CORE-9263: prevent the administrative Grouper user from being added t…

slr71:
…o or removed from any teams

I'm going to be evil and merge this change immediately as well. Reviews are still welcome.


slr71:
Heh, I found a couple of bugs and fixed them. Hopefully they were the only ones. 😬

PR #68, opened 2018-08-15
CORE-8430: added the `start_date` and `end_date` parameters to the ap…

slr71:
…p search endpoints


slr71:
Thanks for the review!

Issue #71, opened 2018-10-10
Add `/communities` endpoints for App Community CRUD and admin management


psarando:
These endpoints will mostly be the same as the `/teams` endpoints, but will operate on a new `communities` Grouper folder, and communities will not require members apart from admins.

The tentative list of endpoints to be added:
* `GET /communities`
* `POST /communities`
* `GET /communities/:name`
* `PATCH /communities/:name`
* `DELETE /communities/:name`
* `GET /communities/:name/admins`
* `POST /communities/:name/admins`
* `POST /communities/:name/admins/deleter`
* `GET /communities/:name/privileges`
* `POST /communities/:name/privileges`

ianmcorvidae:
Gonna be the one to ask the annoying procedural question: should we make a small (go, probably) microservice that terrain delegates to for this group of endpoints?

psarando:
> should we make a small (go, probably) microservice that terrain delegates to for this group of endpoints?

Not at this time. All the current `/teams` endpoints are already implemented in terrain, and these endpoints are just going to reuse those existing functions, but with minor tweaks, such as Grouper folder name changes.

PR #72, opened 2018-10-11
Add `/communities` endpoints for App Community management.

psarando:
This PR will add `/communities` endpoints for App Community CRUD and admin management for #71.

These changes mainly modify the existing teams functions for creating groups as either `iplant:de:<env>:teams:<user>:<name>` or `iplant:de:<env>:communities:<name>`.

New Communities will be publicly viewable by default, with the creator as the initial admin, but an endpoint is available for adding or removing additional admins.

### DIFF
--- a/src/terrain/clients/iplant_groups.clj
+++ b/src/terrain/clients/iplant_groups.clj
@@ -299,23 +299,28 @@
   (get-team user name)
   nil)
 
-(defn- update-team* [team-type user name updates]
+(defn update-team [user name updates]
   (let [client  (get-client)
-        folder  (get-team-folder-name client team-type)
+        folder  (get-team-folder-name client group-type-teams)
         creator (first (string/split name #":" 2))
         group   (full-group-name name folder)]
     (verify-group-exists client user group)
     (->> (update (select-keys updates [:name :description]) :name
-                 full-group-name (get-team-folder-name client team-type creator))
+                 full-group-name (get-team-folder-name client group-type-teams creator))
          (remove-vals nil?)
          (c/update-group client user group)
          (format-group folder))))
 
-(defn update-team [user name updates]
-  (update-team* group-type-teams user name updates))
-
 (defn update-community [user name updates]
-  (update-team* group-type-communities user name updates))
+  (let [client  (get-client)
+        folder  (get-team-folder-name client group-type-communities)
+        group   (full-group-name name folder)]
+    (verify-group-exists client user group)
+    (->> (update (select-keys updates [:name :description]) :name
+                 full-group-name (get-team-folder-name client group-type-communities))
+         (remove-vals nil?)
+         (c/update-group client user group)
+         (format-group folder))))
 
 (defn- delete-team* [team-type user name]
   (let [client (get-client)
--- a/src/terrain/clients/iplant_groups.clj
+++ b/src/terrain/clients/iplant_groups.clj
@@ -225,9 +225,9 @@
 
 ;; Team & Community Functions
 
-(defn- ensure-team-folder-exists [client team-type user]
-  (ensure-folder-exists client (config/grouper-user) (get-team-folder-name client team-type))
-  (ensure-folder-exists client user (get-team-folder-name client team-type user)))
+(defn- ensure-team-folder-exists [client user]
+  (ensure-folder-exists client (config/grouper-user) (get-team-folder-name client group-type-teams))
+  (ensure-folder-exists client user (get-team-folder-name client group-type-teams user)))
 
 (defn- find-teams* [client team-type user search-folder lookup-fn]
   (let [folder (get-team-folder-name client team-type)]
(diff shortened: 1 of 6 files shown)


src/terrain/clients/notifications.clj
```diff
@@ -79,9 +82,22 @@
                       :subject "Added to team"
                       :email true
                       :email_template "added_to_team"
-                      :payload {:email_address (:email user)
+                      :payload {:action "added_to_team"
```
psarando: I also added a `payload.action` to each Team notification, since the UI seems to be looking for an `added_to_team` action, but it's currently checking the `email_template` instead.

src/terrain/clients/notifications.clj
```diff
@@ -79,9 +82,22 @@
...
+                                :email_address (:email user)
                                 :team_name team-name}}))
 
+(defn send-community-admin-add-notification
+  [user team-name]
+  (send-notification {:type           "team"
```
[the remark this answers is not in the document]
ash3rz: I think with the direction we're going, it isn't necessary to have a new notification type for communities. I can update the UI to only open something teams-related depending on the action.

psarando:
> My only comment is that it was a little confusing for me since I kind of viewed it as collaborator lists, teams, and communities are all their own types of groups (maybe because of the folder structure?) and this made it seem to me like a community is a type of team.

I agree that conceptually (and from the users and UI perspective) these should all be considered different types of groups.

In fact, I was hoping to be able to name communities like `iplant:de:<env>:communities:<name>` without having the user's name in there as well, but when I attempted this I got some Grouper error responses.

Since I reused the existing teams functions to manage community groups, and the logic in the collaborator list functions are a bit different than the teams functions, then from the code's perspective the community groups are currently like a different type of team.

slr71:
> In fact, I was hoping to be able to name communities like `iplant:de:<env>:communities:<name>` without having the user's name in there as well, but when I attempted this I got some Grouper error responses.

Those Grouper error responses were probably permissions errors on the folders. Creating the folder and groups using the Grouper admin user should fix that (assuming I understand the problem correctly).

src/terrain/clients/iplant_groups.clj
```diff
@@ -241,75 +245,112 @@
...
-               :else  (partial c/find-groups client user))
-         (get-teams* client user folder))))
+               :else (partial c/find-groups client user))
+         (find-teams* client team-type user folder))))
+
+(defn get-teams [user params]
+  (get-teams* group-type-teams user params))
+
+(defn get-communities [user params]
+  (get-teams* group-type-communities user params))
```
slr71: 👍

src/terrain/clients/iplant_groups.clj
```diff
@@ -241,75 +245,112 @@
...
       (grant-initial-team-privileges client user full-name public_privileges)
-      (format-group (get-team-folder-name client) group))))
+      (format-group (get-team-folder-name client team-type) group))))
 
-(defn get-team [user name]
+(defn add-team [user request]
+  (add-team* group-type-teams user request))
+
+(defn add-community [user request]
+  (add-team* group-type-communities user request))
```
slr71: 👍

src/terrain/clients/iplant_groups.clj
```diff
@@ -324,29 +365,50 @@
...
-    (grant-member-privileges client user group members)
+    (grant-privileges-fn client user group members)
     (c/add-group-members client user group members)))
 
-(defn remove-team-members [user name members]
+(defn add-team-members [user name members]
+  (add-team-members* group-type-teams grant-member-privileges user name members))
+
+(defn add-community-admins [user name members]
+  (add-team-members* group-type-communities grant-admin-privileges user name members))
```
slr71: Higher-order functions almost always make life easier. :)

psarando:
> > In fact, I was hoping to be able to name communities like `iplant:de:<env>:communities:<name>` without having the user's name in there as well, but when I attempted this I got some Grouper error responses.
> 
> Those Grouper error responses were probably permissions errors on the folders. Creating the folder and groups using the Grouper admin user should fix that (assuming I understand the problem correctly).

I think you may be correct. I might have been creating the folder with the Grouper admin user, but trying to still create the group with the requesting user. I'll do some more testing to see if I can get the `communities` folder into this PR after all.

src/terrain/clients/iplant_groups.clj
```diff
@@ -241,75 +245,115 @@
...
+      (format-group (get-team-folder-name client group-type-teams) group))))
 
-(defn get-team [user name]
+(defn add-community [user {:keys [name description public_privileges] :or {public_privileges []}}]
   (let [client (get-client)
-        folder (get-team-folder-name client)]
+        folder (get-team-folder-name client group-type-communities)]
+    (ensure-folder-exists client (config/grouper-user) folder)
+    (let [full-name (full-group-name name folder)
+          group     (c/add-group client (config/grouper-user) full-name team-group-type description)]
```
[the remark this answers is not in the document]
psarando: I just tried it, and apparently the groups service will return a `500 Server Error`:
```
{
  "error_code": "ERR_REQUEST_FAILED",
  "grouper_result_code": "PROBLEM_SAVING_GROUPS",
  "grouper_result_message": "There were 0 successes and 1 failures of saving groups."
}
```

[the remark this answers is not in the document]
ianmcorvidae: Ah, okay. So it'll prevent creating it, but only via an obscure Grouper error, heh. I don't know if it's worth checking whether a community of a given name already exists and throwing a more descriptive error in that case.

src/terrain/clients/iplant_groups.clj
```diff
@@ -259,25 +259,28 @@
...
-(defn add-community [user request]
-  (add-team* group-type-communities user request))
+(defn add-community [user {:keys [name description public_privileges] :or {public_privileges []}}]
+  (let [client (get-client)
+        folder (get-team-folder-name client group-type-communities)]
+    (ensure-folder-exists client (config/grouper-user) folder)
+    (let [full-name (full-group-name name folder)
+          group     (c/add-group client (config/grouper-user) full-name team-group-type description)]
+      (grant-initial-team-privileges client (config/grouper-user) full-name user public_privileges)
+      (format-group folder group))))
```
slr71: Other than the point that Ian brought up, my only concern would be that we'd need to make sure that users can rename and delete communities that they administer. I don't think that Grouper privileges will prevent that, but I don't remember for certain.

psarando: Good point. I just tested it, and it seems that admins can delete the community, but not rename it 😕

psarando: Oh wait, it was just a bug in reusing the `update-teams` code (trying to parse and reinsert the creator's name into the group name). It looks like an admin can rename the group after all. 👍

slr71: Ah! Good news!

src/terrain/clients/iplant_groups.clj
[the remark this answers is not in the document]
psarando: That's probably a good idea.

Since the same issue exists for teams (though it would have to be a user duplicating one of their own teams), I think I'll save the fix for both endpoints for another PR.

psarando:
Thanks for the reviews!

PR #73, opened 2018-10-22
Add `/admin/communities` endpoints for admin App Community management.

psarando:
This PR will add `/admin/communities` endpoints for admin App Community management for #71.

These admin endpoints just reuse the regular functions used by the `/communities` endpoints, except they act as the DE grouper user to ensure they always have admin permissions for each action.

The 1 exception to this is `POST /admin/communities`, which will still add the new community as the requesting user, automatically adding them as an admin to that community.


psarando:
Thanks for the reviews! 👍

PR #74, opened 2018-11-21
Check for tagged apps when renaming a community.

psarando:
This PR will update `PATCH /(admin/)communities/{name}` to check for apps tagged in that community when the admin is renaming that community, by adding `retag-apps` and `force-rename` params to these endpoints, so that when these params are omitted (or set to `false`) when renaming the community, an `ERR_EXISTS` will be thrown if any apps are tagged. Otherwise, any app AVU tags will be updated with the new community name if the `retag-apps` param is set to `true`.

This PR also updates some `terrain.clients.metadata.raw` functions with cyverse-de/metadata-client@3.0.1 and an optional `terrain.communities.metadata.attr` config.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -47,6 +47,7 @@
                  [org.cyverse/common-cli "2.8.1"]
                  [org.cyverse/kameleon "3.0.1"]
                  [org.cyverse/heuristomancer "2.8.0"]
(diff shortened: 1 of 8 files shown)


psarando:
Thanks for the reviews! 👍

PR #75, opened 2018-11-30
Add `/communities/:name/members`, `join`, and `leave` endpoints.

psarando:
This PR will add `/communities/:name/members`, `/communities/:name/join`, and `/communities/:name/leave` endpoints.

The `/communities/:name` endpoint already supports the `?member=username` query param.

The current plan for these endpoints is to allow a user to "favorite" a community by becoming a member, and have only their "favorite" communities listed in the Apps window by the UI (via the `/communities/:name?member=username` endpoint).


psarando:
Thanks for the reviews! 👍

PR #76, opened 2018-12-03
add admin endpoint support for OAuth authentication

slr71:


slr71:
Thanks for the review!

PR #77, opened 2019-01-08
moved all route declarations from terrain.core to terrain.routes

slr71:
There will be more changes to follow, but this change shouldn't modify terrain's behavior at all. I wanted to get this into QA so that we can catch any bugs as soon as possible.


slr71:
Thanks for the review!

PR #79, opened 2019-01-16
Add public optin privileges when creating a community

ash3rz:
Was wondering why I couldn't join any communities where I wasn't an admin 😄 I _think_ this is the only spot where it's needed.


ash3rz:
Thanks for the review!

PR #81, opened 2019-02-12
CORE-4: add Swagger docs to `terrain.routes.collaborator`

slr71:
The following changes were made:

- The latest versions of `cyverse-groups-client` and `common-swagger-api` are now being used.
- Some unused endpoints were removed.
- The context path, `/terrain`, is now added to rather than removed from URL paths by middleware.
- Some error handling in `terrain.util.oauth` was modified to provide errors messages that are actually useful.
- Swagger docs were added to the remaining endpoints in `terrain.routes.collaborator`.


src/terrain/routes/schemas/collaborator.clj
```diff
@@ -0,0 +1,124 @@
...
+
+(defschema AddCommunityRequest
+  (select-keys (group-schema/base-group "community") [:name (optional-key :description)]))
+
+(defschema UpdateCommunityRequest
+  (select-keys (group-schema/group-update "community") (map optional-key [:name :description])))
+
+(defschema UpdateCommunityParams
+  {(optional-key :retag-apps)
+   (describe Boolean "Set to `true` to cause apps that are associated with the community to be retagged")
```
psarando: This param only applies when a community is getting renamed. So how about this?
```suggestion
   (describe Boolean "Set to `true` to cause apps that are associated with a renamed community to be retagged")
```

src/terrain/routes/schemas/collaborator.clj
```diff
@@ -0,0 +1,124 @@
+(ns terrain.routes.schemas.collaborator
+  (:use [common-swagger-api.schema :only [describe NonBlankString]]
+        [schema.core :only [defschema optional-key]])
+  (:require [common-swagger-api.schema.groups :as group-schema]
+            [common-swagger-api.schema.subjects :as subject-schema]))
+
+(defn group-member [group-descriptor member-descriptor]
```
psarando: 👍

src/terrain/routes/schemas/collaborator.clj
```diff
@@ -0,0 +1,124 @@
...
+
+;; Team Schemas
+
+(def TeamNamePathParam
+  (describe NonBlankString "The name of the team, including the username prefix (e.g. `username:team-name`)"))
+
+(def TeamRequesterPathParam
+  (describe NonBlankString "The username of the person requesting to join the team"))
+
+(defn team-listing-params [descriptor plural-descriptor]
```
psarando: 👍

src/terrain/routes/token.clj
```diff
@@ -41,5 +14,5 @@
      (GET "/" [:as {{:strs [authorization]} :headers}]
        :summary "Obtain OAuth Tokens"
        :return AccessTokenResponse
-       :description token-route-description
+       :description-file "docs/get-token.md"
```
psarando: 🎉 🎊 👍

src/terrain/services/collaborator_lists.clj
```diff
@@ -28,7 +28,7 @@
     list))
 
 (defn delete-collaborator-list [{user :shortUsername} name {retain-permissions? :retain-permissions}]
-  (if (Boolean/parseBoolean retain-permissions?)
+  (if retain-permissions?
```
psarando: Another bonus of `compojure-api` is proper param validation and coercion 👍

src/terrain/util.clj
```diff
@@ -13,6 +15,10 @@
         (vector? p) p
         :else       [p]))
 
+(defmethod compojure.api.meta/restructure-param :description-file
```
psarando: 🏆 

Maybe this should eventually be defined in the `common-swagger-api` lib, if that's possible? So markdown file docs could be shared between common `terrain` and `apps` service endpoints, for example? Not sure yet how the resource path for a doc in the lib would look 🤔

slr71: I'm pretty sure it'll work without modification if the file happens to be in a different JAR file. We'll just have to watch out for file name collisions.

slr71: I really like the idea of putting this in `common-swagger-api`

slr71:
Thanks for the review!

src/terrain/util.clj
psarando: > We'll just have to watch out for file name collisions.

Good point. I guess we could keep the docs in subfolders that mimic the endpoint or Swagger tag. For example: `resources/docs/apps/app-metadata/get-app-metatada.md` or something like that.

slr71: Yeah, that's a good idea.

PR #82, opened 2019-02-14
CORE-4: added some middleware to remove a query parameter that the DE…

slr71:
… UI always sends but isn't required for most endpoints

### DIFF
--- a/src/terrain/routes.clj
+++ b/src/terrain/routes.clj
@@ -190,14 +190,13 @@
                                      {:name "token", :description "OAuth Tokens"}]
                :securityDefinitions security-definitions}})
   (middleware
-   [wrap-keyword-params
+   [[wrap-query-param-remover "ip-address" #{#"^/terrain/secured/bootstrap"}]
+    wrap-query-params
+    wrap-lcase-params
+    wrap-keyword-params
     clean-context]
    (context "/terrain" []
      (terrain-routes))))
 
 (def app-wrapper
-  (-> app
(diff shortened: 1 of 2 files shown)


src/terrain/routes.clj
```diff
@@ -191,10 +191,13 @@
...
-    wrap-lcase-params
-    wrap-query-params
     clean-context]
    (context "/terrain" []
      (terrain-routes))))
 
-(def app-wrapper (wrap-context-path-adder app "/terrain"))
+(def app-wrapper
+  (-> app
+      (wrap-query-param-remover "ip-address" #{#"^/terrain/secured/bootstrap$"})
```
psarando: Any reason why `wrap-query-param-remover` and `wrap-context-path-adder` can't go into the `middleware` wrapper above?

Especially since it seems these middleware wrappers are applied "backwards" here (I think), so that `wrap-*-params` are applied to requests before `wrap-context-path-adder`, then finally `wrap-query-param-remover`?

I like that the order wrappers are listed in the `middleware` wrapper above is the same order they are applied to requests.

slr71: That's a good point. `context-path-adder` needs to go outside the app definition because of the way that the Swagger UI is handled. The others may be able to go inside it, though. I'll give it a try.

slr71: I couldn't quite get it to work without modifying the middleware, but a small middleware change was enough to get it to work.

slr71:
Thanks for the review!

PR #85, opened 2019-02-26
CORE-107: make the sort-col and sort-dir parameters for the `GET /sec…

slr71:
…ured/favorites/filesystem` endpoint case-insensitive


slr71:
Thanks for the review!

PR #86, opened 2019-03-06
Analyses facade

johnworth:
Adds the secured facade endpoints for the new analyses service. All of the endpoints are swaggerized.

A new configuration option `terrain.analyses.base-uri` has been added with a sane default that should work inside our kubernetes clusters.

### DIFF
--- a/src/terrain/clients/analyses.clj
+++ b/src/terrain/clients/analyses.clj
@@ -26,13 +26,11 @@
   (:body (http/put (analyses-url ["badges" id])
                    {:content-type  :json
                     :as            :json
-                    :form-params   submission-info})
-         {:id id :submission submission-info}))
+                    :form-params   submission-info})))
 
 (defn add-badge
   [submission-info]
   (:body (http/post (analyses-url ["badges"])
                     {:content-type :json
                      :as           :json
-                     :form-params  submission-info})
-         {:submission submission-info}))
+                     :form-params  submission-info})))
--- a/src/terrain/clients/analyses.clj
+++ b/src/terrain/clients/analyses.clj
@@ -2,8 +2,7 @@
   (:require [terrain.util.config :refer [analyses-base-uri]]
             [clj-http.client :as http]
             [cemerick.url :refer [url]]
-            [terrain.auth.user-attributes :refer [current-user]]
-            [slingshot.slingshot :refer [throw+]]))
+            [terrain.auth.user-attributes :refer [current-user]]))
 
 
 (defn analyses-url
--- a/src/terrain/clients/analyses.clj
+++ b/src/terrain/clients/analyses.clj
@@ -9,7 +9,7 @@
   ([components]
    (analyses-url components {}))
   ([components query]
-   (-> (apply url (cons (analyses-base-uri) components))
+   (-> (apply url (analyses-base-uri) components)
        (assoc :query (assoc query :user (:username current-user)))
        (str))))
 
@@ -19,8 +19,7 @@
 
 (defn delete-badge
   [id]
-  (http/delete (analyses-url ["badges" id]) {:as :json}) {:id id}
-  {:id id})
+  (:body (http/delete (analyses-url ["badges" id]) {:as :json})) {:id id})
 
 (defn update-badge
   [id submission-info]
--- a/src/terrain/clients/analyses.clj
+++ b/src/terrain/clients/analyses.clj
@@ -1,0 +1,59 @@
+(ns terrain.clients.analyses
+  (:require [terrain.util.config :refer [analyses-base-uri]]
+            [terrain.util.transformers :refer [add-current-user-to-map]]
+            [clj-http.client :as http]
+            [clojure-commons.client :refer [build-url-with-query]]
+            [clojure-commons.error-codes :refer [ERR_NOT_FOUND
+                                                 ERR_BAD_REQUEST
+                                                 ERR_UNCHECKED_EXCEPTION]]
+            [slingshot.slingshot :refer [throw+]]))
+
+
+(defn analyses-url
+  ([relative-url]
+   (analyses-url relative-url {}))
+  ([relative-url query]
+   (build-url-with-query (analyses-base-uri)
+                         (add-current-user-to-map query) relative-url)))
+
+(defn process-response
+  [resp err-data]
+  (cond
+    (= (:status resp) 404)
+    (throw+ (merge {:error_code ERR_NOT_FOUND} err-data {:msg (:body resp)}))
+
+    (= (:status resp) 400)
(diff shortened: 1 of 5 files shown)


src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,74 @@
...
+    (throw+ (merge {:error_code ERR_BAD_REQUEST} err-data {:msg (:body resp)}))
+
+    (= (:status resp) 500)
+    (throw+ (merge {:error_code ERR_UNCHECKED_EXCEPTION} err-data {:msg (:body resp)}))
+
+    (not (<= 200 (:status resp) 299))
+    (throw+ (merge {:error_code ERR_UNCHECKED_EXCEPTION} err-data {:msg (:body resp)}))
+
+    :else
+    (:body resp)))
```
slr71: `clj-http` throws exceptions for many HTTP response codes by default. It's possible that this function isn't being used unless the request succeeds. I like the way that you're merging in the error data, though. That's a good idea.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,74 @@
...
+
+(defn get-badge
+  [id]
+  (let [u (analyses-url ["badges" id])]
+    (log/info u)
+    (process-response (http/get u {:as :json}) {:id id})))
```
slr71: I really like `clj-http`'s JSON coercion.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,74 @@
...
+  (process-response (http/delete (analyses-url ["badges" id]) {:as :json}) {:id id})
+  {:id id})
+
+(defn update-badge
+  [id submission-info]
+  (process-response (http/put (analyses-url ["badges" id])
+                              {:content-type  :json
+                               :as            :json
+                               :body          (generate-string submission-info)})
+                    {:id id :submission submission-info}))
```
slr71: `clj-http` has request body coercion based on content type too, so you don't have to explicitly do the JSON serialization if you don't want to. See the examples in the [clj-http documentation](https://github.com/dakrone/clj-http#post) for details. Basically, you'd just replace the `:body` key with `:form-fields`. As long as the content type is `:json`, `clj-http` will serialize it for you.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,74 @@
...
+                                                 ERR_BAD_REQUEST
+                                                 ERR_UNCHECKED_EXCEPTION]]
+            [cemerick.url :refer [url]]
+            [medley.core :refer [mapply]]
+            [slingshot.slingshot :refer [throw+]]))
+
+(defn- fix-user
+  [{:keys [user] :as user-map}]
+  (-> user-map
+    (assoc :user (str user "@iplantcollaborative.org"))
```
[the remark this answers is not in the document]
johnworth: Cool, didn't know about either of those. I'll give the latter a shot.

src/terrain/clients/analyses.clj
johnworth: I was wondering about that, but saw a similar function elsewhere in the terrain code. I'll try removing the calls and see what happens.

src/terrain/clients/analyses.clj
johnworth: I'll take a look, thanks.

src/terrain/clients/analyses.clj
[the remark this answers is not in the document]
johnworth: This should be fixed now.

src/terrain/clients/analyses.clj
johnworth: This should be fixed now.

src/terrain/clients/analyses.clj
johnworth: This should be fixed now.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,40 @@
+(ns terrain.clients.analyses
+  (:require [terrain.util.config :refer [analyses-base-uri]]
+            [clj-http.client :as http]
+            [cemerick.url :refer [url]]
+            [terrain.auth.user-attributes :refer [current-user]]
+            [slingshot.slingshot :refer [throw+]]))
```
slr71: I think that `[slingshot.slingshot :refer [throw+]]` can be removed now unless I'm missing something. I didn't see `throw+` used anywhere.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,40 @@
...
+
+
+(defn analyses-url
+  ([components]
+   (analyses-url components {}))
+  ([components query]
+   (-> (apply url (cons (analyses-base-uri) components))
```
slr71: I didn't catch this before, but `(apply url (cons (analyses-base-url) components))` can be shortened to `(apply url (analyses-base-url) components)`. This is optional, of course. I'm just lazy and I usually use whichever option requires less typing. 😆

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,40 @@
...
+       (assoc :query (assoc query :user (:username current-user)))
+       (str))))
+
+(defn get-badge
+  [id]
+  (:body (http/get (analyses-url ["badges" id]) {:as :json})))
+
+(defn delete-badge
+  [id]
+  (http/delete (analyses-url ["badges" id]) {:as :json}) {:id id}
```
slr71: I could be wrong, but I think that the `{:id id}` form on the end of this line is extraneous.

src/terrain/clients/analyses.clj
```diff
@@ -0,0 +1,40 @@
...
+  ([components]
+   (analyses-url components {}))
+  ([components query]
+   (-> (apply url (cons (analyses-base-uri) components))
+       (assoc :query (assoc query :user (:username current-user)))
+       (str))))
+
+(defn get-badge
+  [id]
+  (:body (http/get (analyses-url ["badges" id]) {:as :json})))
```
slr71: 👍

src/terrain/clients/analyses.clj
johnworth: This should be fixed now.

src/terrain/clients/analyses.clj
johnworth: Yeah, it's a hold-over from similar code in the analyses service. I'll just return the body of of the deletion response.

johnworth: This should now be dealt with.

johnworth: Actually, I was wrong. It a hold over from the parse-response parameters. I've nuked it.

PR #88, opened 2019-03-22
CORE-11 Add /apps/elements swagger docs.

psarando:
Added `terrain.routes.apps.elements`, which is basically the same as `apps.routes.apps.elements`, except the potentially obsolete `GET /apps/elements/tools` endpoint has been omitted for now.


psarando:
I'm going to merge this now, but reviews are still welcome 👍

PR #89, opened 2019-03-26
Remove type as req key from AddCollaboratorListRequest

ash3rz:
I _think_ this will fix the issue with creating collaborator lists.  I tried testing it in swagger with my local changes and seemed ok.


ash3rz:
Woot!  Thanks for the review!

PR #90, opened 2019-03-27
CORE-11 Add some /apps endpoint swagger docs.

psarando:
Added swagger docs for `/apps/shredder`, `permission-lister`, `sharing`, and `unsharing` endpoints.

Note that this PR can't be merged until the `common-swagger-api.schema.badges` schemas are replaced with `common-swagger-api.schema.quicklaunches` schemas from `common-swagger-api` v2.10.3.

Then the changes in cyverse-de/common-swagger-api#14 can be versioned on top of `common-swagger-api` v2.10.3, and this PR will be updated with that newer version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -24,8 +24,6 @@
                  [clojurewerkz/elastisch "2.2.0"]
                  [com.cemerick/url "0.1.1" :exclusions [com.cemerick/clojurescript.test]]
                  [compojure "1.5.1"]
-                 [metosin/compojure-api "1.1.8"]   ; should be held to the same version as the one
-                                                   ; used by org.iplantc/clojure-commons
                  [dire "0.5.4"]
                  [me.raynes/fs "1.4.6"]
                  [medley "0.8.4"]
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -39,7 +39,7 @@
                  [org.cyverse/cyverse-groups-client "0.1.7"]
                  [org.cyverse/common-cfg "2.8.1"]
                  [org.cyverse/common-cli "2.8.1"]
-                 [org.cyverse/common-swagger-api "2.10.2"]
+                 [org.cyverse/common-swagger-api "2.10.3-SNAPSHOT"]
```
slr71: I actually just published version `2.10.3`. This may have to be bumped up to `2.10.4-SNAPSHOT`.

psarando: Yep 👍 That's what I was kind of getting at in the PR description: this will have to be bumped to `2.10.4` or higher, depending on when other changes are pushed 😉

slr71: Oh, heh. I probably would have noticed that if I'd actually read the PR description. 😖

psarando: heh, no problem 👍

psarando:
Thanks for the review 👍

PR #91, opened 2019-03-29
CORE-148: added the GET /analyses/:analysis-id/history endpoint

slr71:
I decided not to add Swagger docs to terrain for this endpoint just yet because that would probably involve a switch to compojure-api version 1.1.12, which would take more time.

### DIFF
--- a/src/terrain/routes/metadata.clj
+++ b/src/terrain/routes/metadata.clj
@@ -321,6 +321,9 @@
    (POST "/analyses/shredder" [:as {:keys [body]}]
      (service/success-response (apps/delete-jobs body)))
 
+   (GET "/analyses/:analysis-id/history" [analysis-id]
+     (service/success-response (apps/get-job-history analysis-id)))
+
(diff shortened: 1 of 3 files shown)


src/terrain/routes/metadata.clj
```diff
@@ -331,7 +331,10 @@
      (service/success-response (apps/list-job-steps analysis-id)))
 
    (POST "/analyses/:analysis-id/stop" [analysis-id :as {:keys [params]}]
-     (service/success-response (apps/stop-job analysis-id params)))))
+     (service/success-response (apps/stop-job analysis-id params)))
+
+   (GET "/analyses/:analysis-id/history" [analysis-id]
```
psarando: I might have put this endpoint above `GET /analyses/:analysis-id/parameters`, just to keep the endpoints in alphabetical order, but that's also something that can be done later when this namespace is swaggerized.

slr71: Ah, that seems like a good idea. I'll update it.

slr71:
Thanks for the review!

PR #93, opened 2019-04-04
CORE-8: updated several dependencies

slr71:
This PR doesn't include all of the changes for CORE-8, but I wanted to get it merged as quickly as I could because I'm switching to `compojure-api` version 1.1.12. This switch has been fraught with bizarre dependency issues for some of our other services, so it's best to do as much testing as possible before throwing other changes into the mix.

The good news is that `lein-ring` is working with this set of dependencies. Hopefully this will provide a way to get it working with the metadata service again in the near future. `lein-ring` is really convenient, and I'd hate to give it up.


psarando:
I attempted to run this branch on my box, but I'm getting compiler errors for the swagger docs when running with `lein do clean, ring server-headless` 😞

`clojure.lang.Compiler$CompilerException: Syntax error compiling compojure.core/compile-route at (compojure/api/meta.clj:327:18).`

The swagger docs load when run with `lein run --config /etc/iplant/de/terrain.properties`.

slr71:
Curses! I was afraid it was too good to be true. 😞 I'll try some other configurations and remove support for `lein-ring` if all else fails.

slr71:
Thanks for the review!

PR #95, opened 2019-04-06
CORE-8: switched back to eastisch version 2.2.0

slr71:
Oops! I just realized that `lein eastwood` was reporting arity problems with elastisch version 3.0.0.


slr71:
Thanks for the review!

PR #97, opened 2019-04-10
CORE-11 Add Swagger docs to /apps/categories endpoints

psarando:
This PR will add Swagger docs to `/apps/categories` endpoints with the schemas and docs from cyverse-de/common-swagger-api#17


psarando:
Merging now, but reviews still welcome 👍

PR #100, opened 2019-04-13
CORE-11 Remove `Delegates to metadata service` sections from apps docs

psarando:
This PR will use `common-swagger-api` v2.10.7 doc strings that replaced markdown docs, and `common-swagger-api.schema.ontologies` response schemas added by cyverse-de/common-swagger-api#19.


psarando:
Merging now, but reviews are still welcome 👍

PR #106, opened 2019-04-23
CORE-235 Use org.cyverse/metadata-files v1.0.3 for DOI submissions.

psarando:
This PR will replace the `terrain.clients.datacite` namespace by using `org.cyverse/metadata-files` v1.0.3 to generate DataCite XML from DE metadata, for use in DOI submissions.

This fixes a bug where the previous `terrain.clients.datacite` would generate XML with elements in the same order they were encountered in the DE metadata, but DataCite (and tools like `xmllint`) will validate the XML with a schema like https://schema.datacite.org/meta/kernel-4.1/metadata.xsd, which requires strict ordering of the XML elements.

The `org.cyverse/metadata-files` lib at v1.0.3 can generate DataCite 4.1 XML with elements in the correct order, which will pass DataCite validation.


psarando:
This is a very simple change, so I'll go ahead and merge it now, but reviews are still welcome 👍

PR #110, opened 2019-05-06
CORE-11: switched to a newer version of common-swagger-api

slr71:


slr71:
I'm going to merge this change immediately. Reviews are still welcome, though.

PR #111, opened 2019-05-06
CORE-11 Fixes for URL import and tool search endpoints

psarando:
This PR will:
* Update `terrain.services.metadata.internal-jobs/launch-url-import-job` to use `terrain.clients.apps.raw/submit-job`.
* Remove the obsolete `terrain.clients.apps/submit-job` function.
* Bump `common-swagger-api` dependency to version `2.11.4`, which includes a fix from cyverse-de/common-swagger-api#27 that added paging params to the `/tools` endpoint.


psarando:
Merging now, but reviews are still welcome 🙇

PR #113, opened 2019-05-07
CORE-3: made the `overwrite` and `notify` query parameters optional i…

slr71:
…n the `/coge/genomes/{genome-id}/export-fasta` endpoint


slr71:
Thanks for the review!

PR #115, opened 2019-05-10
CORE-21 Add swagger docs for /webhooks

ash3rz:


ash3rz:
Thanks for the reviews!

PR #30, opened 2017-06-13
Added endpoints to create and list teams.

dennisroberts71:
Example team creation:

```
$ curl -sH "$AUTH_HEADER" -H "Content-Type: application/json" -d '{"name":"aqua-team","description":"mah team description","public_privileges":["view","optin"]}' "http://localhost:8080/teams" | jq .
{
  "description": "mah team description",
  "name": "dennis:aqua-team",
  "type": "group",
  "extension": "aqua-team",
  "id": "230705b9ee8a40fa80c12bc2a1ba2efc",
  "display_extension": "aqua-team",
  "display_name": "iplant:de:de-2:teams:dennis:aqua-team",
  "id_index": "11104"
}
```

Team listings can be filtered by `creator` (the username of the user who created the team), `member` (the username of a member of the group--only groups containing that user will be listed), and `search` (a partial group name search). Grouper doesn't provide a group name search when searching for groups containing a member, so the `search` parameter is approximated in terrain for this case.

PR #116, opened 2019-05-14
CORE-11 Add Swagger docs to bootstrap and logout endpoints.

psarando:
This PR will use the Swagger docs added by cyverse-de/common-swagger-api#30 for the `/bootstrap` and `/logout` endpoints.

Also includes a fix for the `logout` endpoint to prevent terrain's middleware from stripping the required `ip-address` param.


psarando:

PR #118, opened 2019-05-15
Add support for scroll and scroll_id to this facade

ianmcorvidae:
I missed this when reviewing the original PR, but these are also things that can be passed! Scroll sets a timeout, but as a string (e.g. "1m" for one minute), and if a scroll timeout is provided the response will include a scroll_id. Then, subsequent requests can get more results from the same search by passing back the scroll ID. This is pretty much the same as the underlying ES stuff, and it was added for the use case of pulling down the ES representations of a bunch of things at once.

Since in the case of a scroll_id being passed the query isn't, now every key is marked optional, but, alas.


ianmcorvidae:
Cool, thanks for the review!

PR #119, opened 2019-05-18
CORE-11 Add Swagger docs for /reference-genomes endpoints.

psarando:
This PR will add the Swagger docs for `/reference-genomes` endpoints added by cyverse-de/common-swagger-api#32.


psarando:

PR #123, opened 2019-05-31
Remove some obsolete terrain.util.service functions.

psarando:
This PR will update a few hold-out namespaces in terrain to use `clojure-commons.assertions` so that the duplicate assertions in `terrain.util.service` can be removed.

It will also remove other obsolete functions in `terrain.util.service` that are no longer in use.


psarando:
I'll go ahead and merge this now, since it seems fairly safe according to IntelliJ and `lein eastwood`, and it's probably good to get it in before the weekend's QA tests.

PR #125, opened 2019-06-10
CORE-11 Add Swagger docs to /admin/tool-requests endpoints

psarando:
This PR will add the Swagger docs added by cyverse-de/common-swagger-api#40 to `/admin/tool-requests` endpoints.

It will also remove the unused `get-authenticated-user` apps service call, which allows the obsolete `terrain.clients.apps` namespace to be removed.


psarando:

PR #126, opened 2019-06-11
Remove some obsolete util functions.

psarando:
This PR will remove some obsolete functions from `terrain.util.service` and `terrain.util.transformers`.


psarando:
I'll go ahead and merge this now so the weekend QA tests can really hit these changes, but reviews still welcome, especially if anyone wants to keep 1 of these util functions for testing in the repl or anything (`eastwood` and local testing assures me they really aren't used anywhere else).

PR #127, opened 2019-06-20
CORE-9 Add Swagger docs to filesystem navigation endpoints

psarando:
This PR will use the docs and schemas added by cyverse-de/common-swagger-api#41 in the `/filesystem/root` and `/filesystem/directory` endpoints.

These endpoints still need to be moved into a `filesystem` tagged context, so they don't stay in the `default` group of the Swagger UI. So this PR is marked as a `WIP` for now.

### DIFF
--- a/src/terrain/routes/schemas/filesystem/navigation.clj
+++ b/src/terrain/routes/schemas/filesystem/navigation.clj
@@ -1,0 +1,39 @@
+(ns terrain.routes.schemas.filesystem.navigation
+  (:use [common-swagger-api.schema :only [describe NonBlankString]])
+  (:require [common-swagger-api.schema.data.navigation :as nav-schema]
+            [schema.core :as s]
+            [schema-tools.core :as schema-tools]))
+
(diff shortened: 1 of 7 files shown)


src/terrain/routes/schemas/filesystem/navigation.clj
```diff
@@ -0,0 +1,39 @@
...
+
+(s/defschema RootResponses
+  (update-in nav-schema/NavigationRootResponses
+             [200 :schema]
+             merge {:roots [RootListing]}))
+
+(s/defschema DirectoryResponses
+  (update-in nav-schema/NavigationResponses
+             [200]
+             merge {:schema DirectoryResponse}))
```
psarando: Since cyverse-de/common-swagger-api#41 defines the entire `:responses` docs for these endpoints, but `terrain` returns slightly modified responses from these endpoints, then this PR will modify the `200` success response docs with terrain's specifics.

Hopefully this doesn't get too tedious for the rest of the `data-info` docs in `terrain` 😓

psarando:
I'll go ahead and merge this now, but reviews are still welcome.

PR #130, opened 2019-07-22
added the InCommon CA certificate to the trusted CA certificates in t…

slr71:
…he image in order to support external sites that use this certificate authority

NCSU uses this certificate authority, so I'm adding it to our trusted CA certificates in the images that need to access CAS for convenience.


slr71:
Thanks for the review!

PR #131, opened 2019-08-20
Vice logs

johnworth:
Adds initial support for getting information needed to stream the logs from a running VICE app.

### DIFF
--- a/src/terrain/clients/app_exposer.clj
+++ b/src/terrain/clients/app_exposer.clj
@@ -33,31 +33,3 @@
   [analysis-id]
   (let [external-ids (analysis-external-ids analysis-id)]
     {:pods (apply concat (map :pods (map list-pods-by-external-id external-ids)))}))
-
-(defn- external-id->podmap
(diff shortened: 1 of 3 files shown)


src/terrain/clients/app_exposer.clj
```diff
@@ -0,0 +1,35 @@
...
+(defn- analysis-external-ids
+  "Extracts a list of external IDs from the steps associated with the analysis"
+  [analysis-id]
+  (map :external_id (analysis-steps analysis-id)))
+
+(defn list-analysis-pods
+  "Returns a list of pods associated with the analysis"
+  [analysis-id]
+  (let [external-ids (analysis-external-ids analysis-id)]
+    {:pods (apply concat (map :pods (map list-pods-by-external-id external-ids)))}))
```
slr71: Not a big deal, but you could use `mapcat` for convenience here if you want. Something like this should work:

```clojure
(mapcat (comp :pods list-pods-by-external-id) external-ids)
```

PR #132, opened 2019-09-05
CORE-331 Update DOI completion notification message.

psarando:
This PR will update the DOI completion notification message (sent to the requesting user) to include additional user instructions and the DOI's landing page link.

It will also include the created DOI in curator completion email template values (support added in cyverse-de/iplant-email#8).


psarando:
Merging now, but reviews are still welcome.

PR #133, opened 2019-09-14
CORE-9 Add Swagger docs to filesystem ticket endpoints.

psarando:
This PR will add the Swagger docs from cyverse-de/common-swagger-api#49 to the filesystem ticket endpoints.

### DIFF
--- a/src/terrain/routes/filesystem/tickets.clj
+++ b/src/terrain/routes/filesystem/tickets.clj
@@ -1,0 +1,58 @@
+(ns terrain.routes.filesystem.tickets
+  (:use [common-swagger-api.schema]
+        [ring.util.http-response :only [ok]]
+        [terrain.routes.schemas.filesystem.tickets]
+        [terrain.util :only [optional-routes]]
+        [terrain.util.transformers :only [add-current-user-to-map]])
+  (:require [common-swagger-api.schema.data :as data-schema]
+            [common-swagger-api.schema.data.tickets :as schema]
+            [terrain.clients.data-info :as data]
+            [terrain.util.config :as config]))
+
+;; FIXME: The `coerce-public-param` middleware is only for backwards compatibility
+;; and should be removed in the next version of the API
+(defn- public-param->boolean
+  "Sets the `public` key in `params` to `true` if its string value is `1` or `true`,
+   otherwise sets it to `false`."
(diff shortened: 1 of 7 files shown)


src/terrain/routes/filesystem/tickets.clj
```diff
@@ -0,0 +1,58 @@
...
+
+  (optional-routes
+    [config/filesystem-routes-enabled]
+
+    (context
+      "/filesystem" []
+      :tags ["filesystem"]
+
+      (POST "/tickets" []
+            :middleware [coerce-public-param]
```
psarando: The `public` param used to be optional, defaulting to a `false` value passed along to the required `data-info` param. It also used to only pass `true` to `data-info` if terrain's param was set to `1`.

This middleware is for backwards compatibility, and allows either `1` or `true` to set the `data-info` param to `true`, and anything else will pass along a `false`. The Swagger docs on this endpoint will only show `true` or `false` options, though.

src/terrain/routes/schemas/filesystem/tickets.clj
```diff
@@ -0,0 +1,17 @@
...
+            [schema.core :as s]
+            [schema-tools.core :as st]))
+
+;; Convert the keywords in the `mode` param to strings,
+;; so that the correct param format is passed through to the apps service.
+;; Remove `for-job` param.
+;; FIXME: Make `public` param optional (for backwards compatibility).
+(s/defschema AddTicketQueryParams
+  (-> schema/AddTicketQueryParams
+      (st/dissoc :for-job)
```
psarando: I thought the `data-info` query param `for-job` was for internal use (by the `apps` service) only, so I'm not exposing it in the add or delete ticket endpoints in `terrain`.

slr71: I _think_ that's accurate. We can always update it later if we need to.

src/terrain/routes/filesystem/tickets.clj
```diff
@@ -0,0 +1,58 @@
+(ns terrain.routes.filesystem.tickets
```
slr71: I like the idea of splitting this out into a separate namespace.

psarando:
Thanks for the reviews! 👍

PR #134, opened 2019-09-24
App publication requests.

slr71:
This PR adds the app publication requests endpoints to terrain. I also added `lein-cljfmt` as a dependency, but I haven't tried to use it to reformat all of the files yet. I'll do that in a separate PR.


slr71:
Thanks for the reviews!

PR #136, opened 2019-10-08
CORE-556 Fix query params for admin user info

ash3rz:
One more schema validation error fixed - I don't know why I originally added details to the path-params 😕

### DIFF
--- a/src/terrain/routes/user_info.clj
+++ b/src/terrain/routes/user_info.clj
@@ -34,9 +34,9 @@
    (context "/users" []
      :tags ["admin-user-info"]
      (GET "/:username/groups" []
-          :path-params [username :- UsernameParam
-                        details :- DetailsParam]
+          :path-params [username :- UsernameParam]
+          :query [params DetailsParam]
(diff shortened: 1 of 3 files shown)


src/terrain/routes/user_info.clj
```diff
@@ -34,9 +34,9 @@
    (context "/users" []
      :tags ["admin-user-info"]
      (GET "/:username/groups" []
-          :path-params [username :- UsernameParam
-                        details :- DetailsParam]
+          :path-params [username :- UsernameParam]
+          :query [params DetailsParam]
```
psarando: You could also destructure the `details` out of `params` here:
```suggestion
          :query [{:keys [details]} DetailsParam]
```
Then the `list-groups-for-user` function signature wouldn't have to change.

ash3rz: Ah neat, I haven't seen this used anywhere else (at least not yet) - I'll refactor 👍

ash3rz:
Thanks for the review!

PR #139, opened 2019-10-28
Use de-irods to move folders for permanent ID requests

ianmcorvidae:
and, rely on the share function afterwards to ensure permissions for the DOI curators group

This sets it up so we can change data-info to pass a client user when doing moves and renames, as requested by @tedgin


ianmcorvidae:
I'm _fairly_ sure this is the only place we do a move or rename passing a group as the user. My brain is a bit scrambled today though, so we'll see if the QA regression picks anything up with this.

PR #140, opened 2019-11-27
CORE-671: prevent terrain from following redirects returned by the ap…

slr71:
…ps service

Note that the code to disable redirects has been removed from the clients for the notification agent and metadata services. We really only need to worry about it for the apps service.


slr71:
Thanks for the review!

PR #141, opened 2020-03-04
CORE-868: add keycloak support

slr71:
This PR adds support for JWTs generated and signed by Keycloak while retaining support for CAS. Eventually support for CAS can be removed. When that support is removed, the `/token` endpoint will have to be updated to use Keycloak instead.


slr71:
I just realized that I left a WIP commit in the commit history. I'm going to do a force push to remove that soon.

slr71:
The commit history has been fixed.

slr71:
Thanks for the review!

PR #142, opened 2020-03-06
CORE-919 add support for optionally authenticated endpoints.

slr71:


src/terrain/auth/user_attributes.clj
```diff
@@ -156,13 +156,17 @@
...
+   request to be processed. Requests without credentials will be passed to the handler without authentication
+   routes that require authentication can use the require-authentication middleware."
   [handler]
   (wrap-auth-selection [[get-fake-auth           (wrap-fake-auth handler)]
                         [get-de-jwt-assertion    (wrap-de-jwt-auth handler)]
                         [get-wso2-jwt-assertion  (wrap-wso2-jwt-auth handler)]
                         [get-cas-oauth-token     (wrap-cas-oauth handler)]
-                        [get-keycloak-oidc-token (wrap-keycloak-oidc handler)]]))
+                        [get-keycloak-oidc-token (wrap-keycloak-oidc handler)]
+                        [(constantly true)       handler]]))
```
slr71: Note that this means that the handler generated by this call to `wrap-auth-selection` will never return a 401. I briefly considered modifying `wrap-auth-selection` so that handlers generated by it never return a 401. The idea was that we'd force the caller to explicitly handle the case where there's no authentication information in the request. I reconsidered, though. If the case where authentication information isn't present in the request isn't specifically handled and we receive such a request then returning a 401 seems like the right thing to do.

src/terrain/routes.clj
```diff
@@ -64,6 +64,11 @@
         (tc/with-logging-context {:user-info (cheshire/encode user-info)}
           (handler request))))))
 
+(defn optionally-authenticated-routes
+  []
+  (util/flagged-routes
+   (secured-filesystem-routes)))
```
slr71: For the time being, filesystem routes are going to be available both  inside and outside the `/secured` context. When these routes are called outside the `/secured` context, authentication will only be required for routes that need it. When they're called inside the `/secured` context, authentication will always be required. This allows us to keep supporting the existing DE without encountering the paradox of calling a `/secured/...` endpoint with no authentication information. 😆

slr71: Also note that `optionally-authenticated-routes` has no call to `route/not-found`. This allows the request to fall through to `secured-routes-no-context` if none of the optionally authenticated routes match.

src/terrain/services/filesystem/directory.clj
```diff
@@ -146,90 +148,13 @@
...
-
-(with-post-hook! #'do-paged-listing (paths/log-func "do-paged-listing"))
+  [{user :shortUsername} {:keys [path] :as params}]
+  (let [params (dissoc params :path)]
+    (->> (merge params {:user      (or user "anonymous")
+                        :bad-chars (cfg/fs-bad-chars)
+                        :bad-name  (cfg/fs-bad-names)
+                        :bad-path  (bad-paths user)})
+         (data/list-folder-contents path)
+         (format-page user))))
```
slr71: Getting rid of the error handlers has the potential to break a few QA tests, but I think the results will be worth the test breakages. The primary result is that errors returned by `data-info` will be returned to the caller as-is without any alteration. For example, doing a listing for a directory that doesn't exist used to return a response code of `500` along with a message saying that the service call to `data-info` failed:

```
$ curl -sH "$AUTH_HEADER" "https://qa.cyverse.org/terrain/secured/filesystem/paged-directory?path=/iplant/home/sarahrrrr&limit=1000&offset=0" | jq .
{
  "error_code": "ERR_REQUEST_FAILED",
  "message": "GET http://data-info/data/path/iplant/home/sarahrrrr had a service error: internal error"
}
```

Now it will continue to return a status code of 500, but the error will indicate that the path doesn't exist:

```
$ curl -sH "$AUTH_HEADER" "http://localhost:31325/filesystem/paged-directory?path=/iplant/home/sarahrrrr&limit=1000" | jq .
{
  "error_code": "ERR_DOES_NOT_EXIST",
  "path": "/iplant/home/sarahrrrr"
}
```

There were some error handlers designed to catch errors for missing file paths in `terrain`, but they were never being called because `data-info` returns a status code of `500`. It should probably be updated to return a `404` at some point, but this error is better than what we were getting before. Also, modifying `data-info` to return a `404` in that case won't require any changes to terrain.

I also removed a lot of the query parameter validation, primarily because `compojure-api` does a lot of it for us. Any remaining query parameter validation can be handled by `data-info` when it gets the request.

slr71:
I forced pushed a rebase to avoid problems when we merge in this PR.

src/terrain/routes/schemas/filesystem.clj
```diff
@@ -103,3 +104,7 @@
 
 (s/defschema FilteredUuids
   {:filesystem (describe [UUID] "The filtered list of UUIDs")})
+
+(s/defschema FolderListingParams
+  (merge (st/required-keys data-schema/FolderListingParams [:limit])
```
psarando: 👍

src/terrain/routes.clj
slr71: Good point. I'll add a docstring.

slr71:
Thanks for the reviews!

PR #144, opened 2020-03-11
CORE-919: fix some breaking changes to the paged-directory endpoints.

slr71:


src/terrain/routes/schemas/filesystem.clj
```diff
@@ -106,8 +106,11 @@
   {:filesystem (describe [UUID] "The filtered list of UUIDs")})
 
 (s/defschema FolderListingParams
-  (merge (st/required-keys data-schema/FolderListingParams [:limit])
-         {:path (describe String "The path to the folder in the data store.")}))
+  (-> (st/required-keys data-schema/FolderListingParams [:limit])
+      (st/update :sort-field transform-enum name)
+      (st/update :entity-type transform-enum name)
+      (st/assoc :path (describe String "The path to the folder in the data store."))
+      ((fn [m] (st/assoc m (s/optional-key :sort-col) (st/get-in m [:sort-field]))))))
```
psarando: Nice ✨ 
I think that last function wouldn't have to be anonymous if this used `as->` instead of just `->` 🤔

slr71: Ah, you're probably right. I always forget about `as->`. I'll fix that.

slr71:
Thanks for the review!

PR #147, opened 2020-03-31
CORE-919: fixed a couple more query parameter inconsistencies in the …

slr71:
…paged directory endpoint

### DIFF
--- a/src/terrain/routes/filesystem.clj
+++ b/src/terrain/routes/filesystem.clj
@@ -27,7 +27,10 @@
 (defn- fix-sort-col-param [param-value]
   (as-> param-value v
     (string/lower-case v)
-    (if (= v "lastmodified") "datemodified" v)))
+    (cond
+      (= v "id")           "path"
+      (= v "lastmodified") "datemodified"
+      :else                v)))
 
 (defn secured-filesystem-routes
   "The routes for file IO endpoints."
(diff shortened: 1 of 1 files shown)


src/terrain/routes/filesystem.clj
```diff
@@ -27,7 +27,10 @@
 (defn- fix-sort-col-param [param-value]
   (as-> param-value v
     (string/lower-case v)
-    (if (= v "lastmodified") "datemodified" v)))
+    (cond
+      (= v "id")           "path"
```
ash3rz: I'm not aware of a time when the ui sends an `id` sort column, but I could definitely be missing something. I guess even if not, then this change shouldn't hurt anything.

slr71: Ah, thanks for checking on that. The old version of this endpoint that translated the sort column from `id` to `path`, so I thought I should reinsert it here in case it was used anywhere. I think I might leave it for now. We can always remove it later if we need to.

slr71:
Thanks for the reviews!

src/terrain/routes/filesystem.clj
psarando: The UI used to treat `id` the same as `path` before we added UUIDs to file/folder metadata, so this was probably some cruft from those days; but it's also probably fine leaving this check until we retire the old UI.

src/terrain/routes/filesystem.clj
```diff
@@ -27,7 +27,10 @@
...
+      (= v "lastmodified") "datemodified"
+      :else                v)))
```
ianmcorvidae: I know this is already merged but being slightly picky, I'd prefer `condp` here:

```clojure
(condp = v
    "id" "path"
    "lastmodified" "datemodified"
    v)
```

cuts down a little on the repetition of `(= v ...)`

slr71: I have mixed feelings about `condp`, which was the primary reason I didn't use it. I like to have all of the conditions (or condition arguments) aligned on the left and all of the actions (or return values) aligned on the right. My editor will do that for me when I use `cond`, but not when I use `condp`. 😆 I haven't tried to figure out how to get Clojure mode in Emacs to align `condp` the way I like yet, so I generally stick with `cond`. That's not the greatest reason to do so, but it's the only reason I have at the moment. 🤣

PR #148, opened 2020-04-01
turn off response validation and coercion for the paged-directory end…

slr71:
…point


slr71:
Thanks for the reviews!

PR #149, opened 2020-04-07
CORE-971: added the GET /terrain/apps/:system-id/:app-id/listing endp…

slr71:
…oint


slr71:
Thanks for the review!

PR #151, opened 2020-04-08
CORE-675 - requests and concurrent job limit routes

slr71:
This PR adds several routes that can be used to allow users to submit VICE access requests and view the requests that they've submitted. It also adds administrative routes for requests and concurrent job limits.

There's a bit of an inconsistency between the way that the administrative and non-administrative request routes are defined. The non-administrative routes are explicitly defined for each type of request (currently, there's only one type of request: `vice`). The administrative routes are written to handle multiple different types of requests instead. The reason for this is that it's convenient to be able to specify the `detail` fields that are required for each request in the non-administrative endpoints. In the administrative endpoints, the details are never changed, so we can treat the details as mostly opaque and simply display them to administrators who are responding to incoming requests.

Another benefit of allowing administrative endpoints to process multiple types of requests is that it allows us to create a consolidated page in the administrative UI that can be used to process multiple different types of requests. We don't have to have a consolidated page, but I wanted to make it a possibility in case we decide that we want to have one.

### DIFF
--- a/src/terrain/services/requests.clj
+++ b/src/terrain/services/requests.clj
@@ -1,10 +1,12 @@
 (ns terrain.services.requests
-  (:require [terrain.clients.requests :as rc]))
+  (:require [clojure-commons.exception-util :as cxu]
+            [terrain.clients.requests :as rc]))
 
 ;; Request type constants
 (def vice-request-type "vice")
 
 (defn list-vice-requests
+  "Lists VICE requests. for the currently authenticated user."
   [{username :shortUsername}]
   (rc/list-requests {:request-type    vice-request-type
                      :requesting-user username}))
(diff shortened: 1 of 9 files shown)


src/terrain/services/requests.clj
```diff
@@ -0,0 +1,101 @@
...
+(def get-vice-request
+  "Gets information about a VICE request. If the request exists but is not a VICE request then an exception will be
+   thrown to cause the service endpoint to return a 404."
+  (comp (partial validate-request-type vice-request-type) get-request))
+
+(defn validate-request-user
+  "Verifies that the current user is the person who submitted the request. If the user did not submit the request
+   then an exception will be thrown to cause the service endpoint to return a 404. This is useful in cases where
+   the user is getting information about a request, and we don't want them to be able to obtain information about
+   requests submitted by other users."
```
psarando, line 53: 👍

src/terrain/services/requests.clj
```diff
@@ -0,0 +1,101 @@
...
+  [{username :shortUsername} {requesting-user :requesting_user request-id :id :as request}]
+  (when-not (= username requesting-user)
+    (cxu/not-found (str "request " request-id " not found")))
+  request)
+
+(defn- request-update-fn
+  "Returns a function that can be used to update a request with a default update message and a given request
+   status code."
```
psarando, line 61: 👍

slr71:
Thanks for the review!

PR #152, opened 2020-04-10
Ensure limit param gets passed through to backing service

johnworth:
Refactors the dashboard-aggregator client and route so the limit query parameter gets passed down to the dashboard-aggregator service.

Tested locally, both logged in and logged out. Appears to be working as intended.

### DIFF
--- a/src/terrain/clients/dashboard_aggregator.clj
+++ b/src/terrain/clients/dashboard_aggregator.clj
@@ -8,11 +8,17 @@
               [terrain.util.config :as config]))
 
 (defn- dashboard-aggregator-url
-  [& parts]
-  (str (apply url (config/dashboard-aggregator-url) parts)))
+  ([limit]
+    (dashboard-aggregator-url [] limit))
+
+  ([components limit]
+    (-> (apply url (config/dashboard-aggregator-url) components)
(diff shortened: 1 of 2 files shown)


src/terrain/clients/dashboard_aggregator.clj
```diff
@@ -8,11 +8,17 @@
...
 (defn- dashboard-aggregator-url
-  [& parts]
-  (str (apply url (config/dashboard-aggregator-url) parts)))
+  ([limit]
+    (dashboard-aggregator-url [] limit))
+
+  ([components limit]
+    (-> (apply url (config/dashboard-aggregator-url) components)
+        (assoc :query {:limit limit})
+        (str))))
```
slr71, line 17: Passing the query parameters to `cemerick.url/url` certainly works, and I haven't mentioned this in previous PRs just because this works well. I've found it a little more convenient to use the `:query-params` option in `clj-http`, though. You can find an example here: https://github.com/cyverse-de/terrain/blob/master/src/terrain/clients/requests.clj#L25-L34.

I don't have a strong preference for either option. I just wanted to mention this in case it turns out to be helpful. Feel free to use whichever option you prefer.

PR #155, opened 2020-06-04
CORE-19 Add swagger docs to tag routes

ash3rz:


src/terrain/routes/tags.clj
```diff
@@ -10,32 +13,76 @@
...
+        :summary schema/PatchTagSummary
+        :description schema/PatchTagDescription
+        :return nil
+        (ok (tags/update-user-tag tag-id body)))
+
+     (DELETE "/user/:tag-id" []
+        :path-params [tag-id :- schema/TagIdPathParam]
+        :summary schema/DeleteTagSummary
+        :description schema/DeleteTagDescription
+        :return nil
```
ash3rz, line 87: The one issue I'm still having is these last 2 endpoints are supposed to return `nil`, but I'm actually getting back Elasticsearch responses.  Any suggestions on how to fix that would be appreciated.

slr71: You can remove the `:return nil` bits and update the handlers to something like this:

``` clojure
    (tags/update-user-tag tag-id body)
    (ok)
```

ash3rz: Oh! I didn't realize it'd be that easy 😊 Thanks!  I'll refactor

src/terrain/clients/metadata/raw.clj
```diff
@@ -162,46 +162,47 @@
...
+              (json-post-options body
+                                 {:data-type data-type
+                                  :type (:type params)}))))
 
 (defn get-tags-by-value
   [contains limit]
-  (http/get (metadata-url "tags" "suggestions") (get-options (remove-nil-values {:contains contains
-                                                                                 :limit limit}))))
+  (:body (http/get (metadata-url "tags" "suggestions") (json-get-options (remove-nil-values {:contains contains
+                                                                                 :limit limit})))))
```
slr71, line 185: This is really minor, but it might help to line up `:limit` with `:contains` to show that it's part of the same map. I think that IntelliJ has an option to line up the code for you. Otherwise, you can run `lein cljfmt fix src/terrain/clients/metadata/raw.clj` to have it reformat the code for you. That will probably make some additional changes as well. If you prefer, you can also run `lein cljfmt check src/terrain/clients/metadata/raw.clj` to have it display the recommended changes.

src/terrain/routes/tags.clj
```diff
@@ -10,32 +13,76 @@
   (util/optional-routes
    [#(and (config/filesystem-routes-enabled) (config/metadata-routes-enabled))]
 
-   (GET "/filesystem/entry/tags" []
-     (tags/list-all-attached-tags))
+   (context "/filesystem/entry" []
+            :tags ["tags"]
```
slr71, line 17: I really like using contexts; they prevent a bunch of duplicate code. 👍

src/terrain/routes/tags.clj
```diff
@@ -10,32 +13,76 @@
...
+        (ok))
 
-   (PATCH "/filesystem/entry/:entry-id/tags" [entry-id type :as {body :body}]
-     (tags/handle-patch-file-tags entry-id type body))
+      (GET "/:entry-id/tags" []
+        :path-params [entry-id :- TargetIdParam]
+        :summary schema/GetAttachedTagSummary
+        :description schema/GetAttachedTagDescription
+        :return schema/TagList
+        (ok (tags/list-attached-tags entry-id)))
```
slr71, line 36: Adding a context here for the `/:entry-id` or even the `/:entry-id/path` URL paths would eliminate the need for the duplicate `:path-params` entries. This code has a good example of that: https://github.com/cyverse-de/terrain/blob/eff4049c1b4ecbee7659ce1d2dc03633c8ebcc2f/src/terrain/routes/analyses.clj#L87-L95

src/terrain/routes/tags.clj
```diff
@@ -10,32 +13,76 @@
...
+        :description schema/PatchTagDescription
+        (tags/update-user-tag tag-id body)
+        (ok))
+
+     (DELETE "/user/:tag-id" []
+        :path-params [tag-id :- schema/TagIdPathParam]
+        :summary schema/DeleteTagSummary
+        :description schema/DeleteTagDescription
+        (tags/delete-user-tag tag-id)
+        (ok)))))
```
slr71, line 88: These two could go in a common context as well.

src/terrain/services/metadata/tags.clj
```diff
@@ -16,7 +16,7 @@
 (defn- update-tags-targets
   [response]
   (letfn [(fmt-tgt ([tgt] (update tgt :id uuidify)))
-          (parse-resp ([response] (-> response :body slurp (json/parse-string true) :tags)))]
+          (parse-resp ([response] (-> response :body (json/parse-string true) :tags)))]
```
psarando, line 19: Since `meta/update-attached-tags ` is already calling `:body` on its response, and it's making the request with `:as :json`, then I don't think the extra `:body` and `json/parse-string` calls are required here.

In fact, it looks like `parse-resp` is no longer needed at all, and `(:tags response)` can be used in the `deseq` below.

src/terrain/services/metadata/tags.clj
```diff
@@ -153,7 +149,7 @@
 (defn delete-all-user-tags
   "Deletes all tags that were created by the authenticated user."
   []
-  (let [tag-ids (map :id (:tags (svc/decode-json (:body (meta/list-user-tags)))))
+  (let [tag-ids (map :id (:tags (:body (meta/list-user-tags))))
```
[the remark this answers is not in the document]
slr71: Ah, good catch! I missed that.

src/terrain/clients/metadata/raw.clj
ash3rz: Ah ok, I'll try that out. I've been lining things up manually because IntelliJ wants to space things out super far (even though it's set to 2 spaces for a tab).

src/terrain/routes/tags.clj
ash3rz: OK, cool - this is a great example.  I wasn't sure how far to go with contexts, like maybe just one level was more than enough, but this looks more useful.  Thanks!

src/terrain/services/metadata/tags.clj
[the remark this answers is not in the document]
psarando: It's bit me enough times now in my swagger migrations that I'm on the lookout for it now 😅

PR #156, opened 2020-06-04
Fix arity errors in the time limit endpoints

johnworth:
When I modified the time-limit endpoints (non-admin), I accidentally broke them by attempting to manually set the username in the requests, rather than just using the shortUsername of the logged in user. This resulted in errors popping up in QA. This PR should fix them.


johnworth:
Thanks for the review.

PR #157, opened 2020-06-04
Add new VICE admin endpoints

johnworth:
This PR changes the VICE time-limit endpoints to use the new admin versions in app-exposer (which takes the analysisID instead of the externalID) along with adding the following VICE admin endpoints
* POST /vice/analyses/:analysisID/save-and-exit
* POST /vice/analyses/:analysisID/exit
* POST /vice/analyses/:analysisID/save-output-files
* POST /vice/analyses/:analysisID/download-input-files
* GET /vice/analyses/:analysisID/external-id

### DIFF
--- a/src/terrain/routes/vice.clj
+++ b/src/terrain/routes/vice.clj
@@ -66,5 +66,12 @@
         :path-params [analysis-id :- vice-schema/AnalysisID]
         :summary "Download input files"
         :description "Downloads input files to the analysis container without changing the status of the analysis"
-        (ok (vice/admin-download-input-files analysis-id))))))
+        (ok (vice/admin-download-input-files analysis-id)))
+        
+      (GET "/analyses/:analysis-id/external-id" []
+        :path-params [analysis-id :- vice-schema/AnalysisID]
+        :return vice-schema/ExternalIDResponse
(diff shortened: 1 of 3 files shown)


src/terrain/routes/vice.clj
```diff
@@ -30,25 +30,48 @@
...
+        :summary "Download input files"
+        :description "Downloads input files to the analysis container without changing the status of the analysis"
+        (ok (vice/admin-download-input-files analysis-id)))
+        
+      (GET "/analyses/:analysis-id/external-id" []
+        :path-params [analysis-id :- vice-schema/AnalysisID]
+        :return vice-schema/ExternalIDResponse
+        :summary "Get external UUID"
+        :description "Returns the external UUID associated with the analysis. VICE analyses only have a single external UUID"
+        (ok (vice/admin-external-id analysis-id))))))
```
slr71, line 76: This isn't a requirement, just something that I like to do. When I have several URL paths with a common base path, I'll typically create another context for it. This is especially useful when one of the URL path elements contains a path parameter. I don't have an extremely strong preference, so pick whichever option you prefer. 😁

johnworth:
Thanks for the review!

PR #158, opened 2020-06-12
Make :containerID and :message fields optional

johnworth:
A couple of the fields in the VICE schema cause issues in the admin UI when they're not optional. This PR should fix that.

### DIFF
--- a/src/terrain/routes/schemas/vice.clj
+++ b/src/terrain/routes/schemas/vice.clj
@@ -46,12 +46,12 @@
      :command (describe [String] "The command used to start the analysis")}))
 
 (defschema ContainerStateWaiting
-  {:reason  (describe (maybe String) "The reason the container is in the waiting state")
-   :message (describe (maybe String) "The message associated with the waiting state")})
+  {:reason                 (describe (maybe String) "The reason the container is in the waiting state")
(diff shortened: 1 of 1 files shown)


src/terrain/routes/schemas/vice.clj
```diff
@@ -46,12 +46,12 @@
      :command (describe [String] "The command used to start the analysis")}))
 
 (defschema ContainerStateWaiting
-  {:reason  (describe (maybe String) "The reason the container is in the waiting state")
-   :message (describe (maybe String) "The message associated with the waiting state")})
+  {:reason                 (describe (maybe String) "The reason the container is in the waiting state")
+   (optional-key :message) (describe (maybe String) "The message associated with the waiting state")})
```
ianmcorvidae, line 50: Probably yes, but just to double-check, is it intended that this has both `optional-key` and `maybe` used? (i.e., can it both be not present or null?)

johnworth: Technically possible, though unlikely. This stuff comes directly from the k8s API and I'm a bit uncertain about the possible combinations that are possible. I've already been bitten by it once, so I made it optional across the board to hopefully prevent a bunch of churn in PRs.

johnworth:
Thanks for the reviews.

PR #159, opened 2020-06-18
CORE-1027: allow administrators to impersonate users for troubleshooting

slr71:
This PR finally adds the ability for administrators to impersonate users without having to run terrain locally and jump through a bunch of hoops. The administrative token endpoints use HTTP basic authentication, and they don't do any authorization checks themselves. Instead, they pass the information to Keycloak and allow Keycloak to do the validation. The drawback to this approach is that it means that we'll have to maintain administrative privileges for impersonation tokens separately through Keycloak. The advantage is that we can be more selective about which users have the ability to impersonate other users.

### DIFF
--- a/src/terrain/clients/keycloak.clj
+++ b/src/terrain/clients/keycloak.clj
@@ -26,3 +26,15 @@
                                    :username      username
                                    :password      password}
                      :as          :json})))
+
+(defn get-impersonation-token
+  "Obtains an impersonation token for troubleshooting purposes."
+  [subject-token username]
+  (:body (http/post (keycloak-url "protocol" "openid-connect" "token")
+                    {:form-params {:grant_type           "urn:ietf:params:oauth:grant-type:token-exchange"
+                                   :client_id            (config/keycloak-client-id)
+                                   :client_secret        (config/keycloak-client-secret)
+                                   :subject_token        subject-token
+                                   :requested_token_type "urn:ietf:params:oauth:token-type:access_token"
+                                   :requested_subject    username}
+                     :as          :json})))
--- a/src/terrain/clients/keycloak.clj
+++ b/src/terrain/clients/keycloak.clj
@@ -6,7 +6,7 @@
 (defn- keycloak-url
   "Builds a Keycloak URL with the given path components."
   [& components]
-  (str (apply curl/url (config/keycloak-base-uri) "auth" "realms" (config/keycloak-realm) components)))
+  (str (apply curl/url (config/keycloak-base-uri) "realms" (config/keycloak-realm) components)))
 
 (defn get-oidc-certs
   "Retrieves a list of active public certificates from Keycloak."
@@ -15,3 +15,14 @@
                 {:as :json})
       :body
       :keys))
+
+(defn get-token
+  "Obtains an authorization token."
+  [username password]
+  (:body (http/post (keycloak-url "protocol" "openid-connect" "token")
+                    {:form-params {:grant_type    "password"
+                                   :client_id     (config/keycloak-client-id)
+                                   :client_secret (config/keycloak-client-secret)
+                                   :username      username
+                                   :password      password}
+                     :as          :json})))
--- a/src/terrain/routes.clj
+++ b/src/terrain/routes.clj
@@ -167,6 +167,7 @@
 (defn unsecured-routes
   []
   (util/flagged-routes
+   (admin-token-routes)
    (token-routes)
    (unsecured-misc-routes)
    (unsecured-notification-routes)))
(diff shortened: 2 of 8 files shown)


src/terrain/clients/keycloak.clj
```diff
@@ -6,7 +6,7 @@
 (defn- keycloak-url
   "Builds a Keycloak URL with the given path components."
   [& components]
-  (str (apply curl/url (config/keycloak-base-uri) "auth" "realms" (config/keycloak-realm) components)))
+  (str (apply curl/url (config/keycloak-base-uri) "realms" (config/keycloak-realm) components)))
```
slr71, line 9: The `auth` component was already in the configuration values files, so I just removed it here. I thought it was better to try to keep the configuration settings consistent between Terrain and Sonora than to avoid the code change.

src/terrain/routes.clj
```diff
@@ -167,6 +167,7 @@
 (defn unsecured-routes
   []
   (util/flagged-routes
+   (admin-token-routes)
```
slr71, line 170: I considered putting `admin-token-routes` under a separate group of routes with a different context, but ultimately decided against it. This works, and I don't think it's overly confusing.

src/terrain/routes.clj
```diff
@@ -247,6 +248,7 @@
                                      {:name "admin-reference-genomes", :description "Admin Reference Genome Endpoints"}
                                      {:name "admin-requests", :description "Admin Request Endpoints"}
                                      {:name "admin-settings", :description "Admin Setting Endpoints"}
+                                     {:name "admin-token", :description "Admin OAuth Tokens"}
```
slr71, line 251: I briefly considered changing the wording to `Admin OAuth and OIDC Tokens`. Ultimately, however, OIDC tokens _are_ OAuth tokens. I figured this was close enough.

src/terrain/routes/token.clj
```diff
@@ -14,4 +14,36 @@
...
+       :summary "Obtain OAuth Tokens"
+       :return AccessTokenResponse
+       :description-file "docs/get-token.md"
+       (oauth/get-cas-token authorization))
+
+     (GET "/keycloak" [:as {{:strs [authorization]} :headers}]
+       :summary "Obtain Keycloak OIDC Tokens"
+       :return AccessTokenResponse
+       :description-file "docs/get-token.md"
+       (oauth/get-keycloak-token authorization)))))
```
slr71, line 29: For the time being, I decided to have a default `token` endpoint as well as individual endpoints for each identity provider that we support. This allows the use case that users are familiar with to keep working while allowing us to obtain access tokens for specific identity providers. The endpoints for individual identity providers will likely go away when we fully migrate to Keycloak.

src/terrain/routes/token.clj
```diff
@@ -14,4 +14,36 @@
...
+       :return AccessTokenResponse
+       :description-file "docs/get-admin-token.md"
+       (oauth/get-admin-token authorization username))
+
+     (GET "/keycloak" [:as {{:strs [authorization]} :headers}]
+       :query [{:keys [username]} AdminKeycloakTokenParams]
+       :summary "Obtain Keycloak OIDC Impersonation Tokens"
+       :return AccessTokenResponse
+       :description-file "docs/get-admin-token.md"
+       (oauth/get-keycloak-admin-token authorization username)))))
```
slr71, line 49: For the time being, we also have a default admin token endpoint as well as an endpoint for the one identity provider that we support that allows us to get impersonation tokens.

Also the admin token endpoints use HTTP basic authentication as well. Because of the way that Keycloak allows users to obtain impersonation tokens, this means that these endpoints have to make two separate API calls: one to get the administrator's token, and another to get the actual impersonation token. I could have made the admin token endpoints require a bearer token instead, but that makes it a bit more cumbersome to use these endpoints. I also could have made the endpoints support both HTTP basic authentication and bearer token authentication, but that would have taken longer, and it's not worth the additional time and effort at the moment.

src/terrain/services/oauth.clj
```diff
@@ -14,7 +15,25 @@
               (String.)
               (string/split #":" 2)))))
 
-(defn get-token [authorization]
+(defn get-cas-token [authorization]
   (if-let [[username password] (get-basic-auth-credentials authorization)]
-    (http-response/ok (client/get-token username password))
+    (http-response/ok (update (client/get-token username password) :expires_in #(Integer/parseInt %)))
```
slr71, line 20: I tried response body coercion for this as well, but this was cleaner.

slr71:
Thanks for the review!

slr71:
> As always, not an expert on any of this stuff, but LGTM :shipit:
> Also learned about `:strs` which I hadn't seen yet (as far as I remember), only `:keys`, so that's cool 👍

Ah, yes, `:strs` is helpful. There's also `:syms` for Clojure symbols, but I'm pretty sure we've never used Clojure symbols as map keys in our code. The Clojure destructuring guide is really helpful if you want to check it out: https://clojure.org/guides/destructuring.

PR #161, opened 2020-06-27
CORE-1045 Add temp workaround for async folder moves in DOI endpoints.

psarando:
This PR will update DOI endpoints to continue processing DOI requests and creation if the source folder is already in the staging or publish directory.

This will allow DOI curators and admins to pre-move data folders into the staging or curated folders before processing DOI requests.


psarando:
> Looks good to me. This might serve as a starting point for adding callbacks via async-tasks as well, since it establishes some idempotence to these endpoints -- would maybe work well for such a callback to just basically be "send this off to be moved, then call this endpoint again to continue the process".

Good idea!

psarando:
Great! Thanks for the reviews! 👍

PR #162, opened 2020-07-15
CORE-1094: made the IP address parameter optional in the bootstrap en…

slr71:
…dpoint


slr71:
Thanks for the review!

PR #163, opened 2020-07-17
CORE-1027: fix a bug that was preventing administrators from acting o…

slr71:
…n behalf of other users


slr71:
> Gotta love those bugs with shadowing values of variables, heh. LGTM

Yep, that was a definite facepalm moment for me. 🤦‍♀️ 

Thanks for the review!

PR #164, opened 2020-07-17
CORE-1097: speed up the permanent ID request listing endpoints.

slr71:
This change updates the permanent ID request listing endpoints to retrieve stat information for the folders corresponding to all of the permanent ID requests at once, which should dramatically improve the responsiveness of the endpoint. In  my local testing, the endpoint took a little over 20 seconds to respond. If we need further responsiveness improvements later, we can add pagination or allow the caller to filter the results.

**Note:** The endpoint to get a single permanent ID request is currently not working. We're getting an error in the metadata service that's preventing the request from succeeding. I've created a Jira issue (CORE-1099) for this. I'm hoping to get this change merged before that problem is fixed, however, because this change should allow the permanent ID request listing to work correctly again in Belphegor.

### DIFF
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -157,17 +157,17 @@
   (log/debug "getting or creating dir: path =" path)
   (let [user (:shortUsername current-user)]
     (cond
-     (not (path-exists? user path))
+      (not (path-exists? user path))
       (create-dir {:user user} {:path path})
 
-     (and (path-exists? user path) (st/path-is-dir? path))
-     path
+      (and (path-exists? user path) (st/path-is-dir? path))
+      path
 
-     (and (path-exists? user path) (not (st/path-is-dir? path)))
-     nil
+      (and (path-exists? user path) (not (st/path-is-dir? path)))
+      nil
 
-     :else
-     nil)))
+      :else
+      nil)))
 
 (defn can-create-dir?
   "Determines if a directory exists or can be created."
@@ -186,7 +186,7 @@
   [{:keys [user]} {:keys [source dest]}]
   (validators/not-superuser user)
   (assertions/assert-valid (= (ft/dirname dest) (ft/dirname source))
-      "The directory names of the source and destination must match for this endpoint.")
+                           "The directory names of the source and destination must match for this endpoint.")
   (let [path-uuid (uuid-for-path user source)]
     (raw/rename user path-uuid (ft/basename dest))))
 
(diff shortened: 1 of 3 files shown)


src/terrain/clients/data_info.clj
```diff
@@ -306,6 +308,32 @@
       (get-in ["ids" (str uuid)])
       walk/keywordize-keys))
 
+(defn ^ISeq stats-by-uuids
+  "Resovles the stat info for the entities with the given UUIDs. The results are not paged.
```
ianmcorvidae, line 312: tpyo

src/terrain/clients/data_info.clj
```diff
@@ -306,6 +308,32 @@
...
+     :ignore-missing      - Ignore non-existent UUIDs.
+     :ignore-inaccessible - Ignore UUIDs referring to items the user lacks permission to view.
+     :filter-include      - Fields to explicitly include in the response.
+     :filter-exclude      - Fields to exclude from the response.
+
+   Returns:
+     It returns a path-stat map containing an additional UUID field."
+  [user uuids params]
+  (let [params (select-keys params [:ignore-missing :ignore-inaccessible :filter-include :filter-exclude])]
+    (-> (http/post (data-info-url "path-info")
```
ianmcorvidae, line 329: I'm not sure if there's a reason this is here, but in general, I think we've been avoiding putting the direct HTTP stuff in this namespace, instead putting it into the `terrain.clients.data-info.raw` namespace and then wrapping it here if there's something useful to do. The function above this one already calls the same endpoint using `terrain.clients.data-info.raw/collect-path-info`, so it'd be nice to add the ignore parameters to that function and use it here rather than duplicating `data-info-url` and the HTTP call.

src/terrain/clients/data_info.clj
slr71: Thanks for catching that.

src/terrain/clients/data_info.clj
slr71: I meant to add a preemptive comment, but you got to the review before I got to it. 😄 I did notice that the direct HTTP stuff was going into `terrain.clients.data-info.raw`, but I put this here for a couple of different reasons.

The first reason was that the few functions that I checked in the `raw` namespace weren't actually parsing the response body and were returning the stream directly. I think I was the first one to create a `raw` namespace in Terrain, and the reason that I did so was because some endpoints were direct pass-through endpoints, and parsing the JSON just to serialize it again was pointless at the time. This function is only being called from orchestration code, so the response body always needs to be parsed. I could have placed this entire function in the `raw` namespace, but doing that would mean that some functions are returning raw response bodies and some aren't. I decided that I'd rather put direct HTTP calls in this namespace rather than put a function that returns a parse response body in the `raw` namespace.

The second reason is that the code seems cleaner to me if I let `clj-http` handle the parsing for me. I could have placed the HTTP call in the raw namespace, called that function from here, and then parsed the response body. In my opinion, however, the resulting code would have been more convoluted. The upshot here is that I chose clarity (in my opinion) over consistency.

My reasons for duplicating `data-info-url` are a little flimsier. This function works a little bit differently from the original, and I went with this option simply because it's the option that I use everywhere. I could have made the original function public and called it instead. That probably would have been a better option. I'll do that instead.

I would rather leave the HTTP call in this namespace, however, if there aren't any strong objections.

ianmcorvidae: Heh, sorry to get to it too quickly. I'm only half-working but I saw a review come up and jumped on it since that's a nice contained parcel of work it's easy to do and then be done with. The reasons you list seem fine to me -- I see that you've made it so `data-info-url` is at least shared, which seems good to me (as an aside, I do like your version more, but it'd be a bigger change to swap all of the raw namespace to that, so probably not worth it right now). We're inconsistent everywhere, so my litmus test, personally, is mostly just "is there a pretty good reason", and seems like there is for what's left!

slr71: No problem for getting to it so quickly. 😄 I appreciate the responsiveness.

slr71:
Thanks for the review!

PR #165, opened 2020-07-20
CORE-1031 Allow logged out users to get root listing and view community data

ash3rz:
The main request from the ticket was to have the root listing return all 4 roots (home, shared with me, community data, trash) even for logged out users.  On top of that, logged out users needed to get a 401 if they tried to navigate to (i.e. get a paged directory listing for) any root other than community data, so that work is also in this PR.

### DIFF
--- a/src/terrain/services/filesystem/directory.clj
+++ b/src/terrain/services/filesystem/directory.clj
@@ -2,7 +2,6 @@
   (:use [clojure-commons.core :only [remove-nil-values]]
         [clojure-commons.validators]
         [kameleon.uuids :only [uuidify]]
-        [ring.util.http-response :only [ok]]
         [slingshot.slingshot :only [try+ throw+]])
   (:require [clojure.tools.logging :as log]
             [clojure.string :as string]
@@ -170,6 +169,6 @@
     (let [params (dissoc params :path)]
       (->> (fix-paged-listing-params listing-user params)
            (data/list-folder-contents path)
-           (format-page user)
-           (ok)))
-    (response/unauthorized "No authentication information found in the request")))
+           (format-page user)))
+    (throw+ {:type :clojure-commons.exception/not-authorized
+             :user user})))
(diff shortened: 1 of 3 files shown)


src/terrain/services/filesystem/directory.clj
```diff
@@ -162,7 +164,12 @@
...
-         (format-page user))))
+  (if-let [listing-user (if (= path (cfg/fs-community-data))
+                          (or user "anonymous")
+                          user)]
+    (let [params (dissoc params :path)]
+      (->> (fix-paged-listing-params listing-user params)
+           (data/list-folder-contents path)
+           (format-page user)
+           (ok)))
+    (response/unauthorized "No authentication information found in the request")))
```
psarando, line 175: I'd keep the `ok` up at the `route` level, since I think of the `ok` and `unauthorized` responses as specific to the routes (imagine if we had to switch from `compojure-api` to something else one of these days).

Our compojure-api routes have exception handling that will catch exceptions and return the appropriate error response. For example, take a look at `terrain.services.filesystem.validators/not-superuser`
https://github.com/cyverse-de/terrain/blob/8cfae8a1f23a3f23d5ae90f12b3c92f03af1061b/src/terrain/services/filesystem/validators.clj#L12-L16
It throws a `clojure-commons.exception/not-authorized` exception type, which gets handled by `clojure-commons.exception/not-authorized-handler`:
https://github.com/cyverse-de/iplant-clojure-commons/blob/99c9d70cbe77be66b3c67a467de0fc887fe93359/src/clojure_commons/exception.clj#L113-L114

So the `ok` could be kept back at the route level, and this function could throw an exception in the `else` part of this `if-let`.

ash3rz: Ah ok, thanks for providing an example!  I'll try this out and refactor

ash3rz:
Thanks for the review!

ianmcorvidae:
I think this may not be working (as discussed elsewhere) because this endpoint comes into `terrain.routes` via `filesystem-navigation-routes`, which does not appear in the `optionally-authenticated-routes` section. Since this endpoint only comes in via `secured-routes`, it ends up subject to `secured-routes-handler` and/or `secured-routes-no-context-handler`, both of which add the `require-authentication` middleware. I'm not completely sure what the right way is to add these into `optionally-authenticated-routes` but I know that @slr71 added that in, so we may need to consult with her on that.

PR #166, opened 2020-07-28
CORE-1031 Add filesystem roots to optionally-authenticated-routes

ash3rz:
This effectively creates GET /filesystem/roots which can be used without auth, but still allows GET /secured/filesystem/roots to work as usual.

This means for Sonora to pick up the changes, it will have to be updated to hit the non-secured route.


slr71:
I'm going to go ahead and merge this one in now.

ash3rz:
Thanks for the review and also for merging!

PR #167, opened 2020-07-28
CORE-1112: modified terrain so that it uses a command-line option rat…

slr71:
…her than environment variables to enable fake authentication

This change eliminates the use of fake authentication environment variables completely. Instead, a username will be passed to `terrain` on the command line using the new `--fake-user` or `-u` option. If this option is specified then fake authentication will be enabled, and the specified username will be used for all requests. To eliminate some of the burden on the developer, it is only necessary to specify the username when fake authentication is enabled. The rest of the required information will be obtained from the `iplant-groups` service. Also, the user attributes will be obtained from the `iplant-groups` service for each request. This adds a tiny bit of overhead to each request when fake authentication is enabled, but it also means that terrain doesn't have to be restarted if, for example, `iplant-groups` isn't available when `terrain` starts.

A new option, `--port` or `-p`, has also been added to make it easier to specify the port number on the command line when using `lein run`. Prior to this change, it would have been necessary to update the terrain configuration file to change the listen port. If the new command-line option is not provided on the command line then the port number will still be obtained from the configuration file.

### DIFF
--- a/src/terrain/auth/user_attributes.clj
+++ b/src/terrain/auth/user_attributes.clj
@@ -1,12 +1,20 @@
 (ns terrain.auth.user-attributes
+  (:use [slingshot.slingshot :only [try+]])
   (:require [clojure.string :as string]
             [clojure.tools.logging :as log]
             [clojure-commons.response :as resp]
             [clojure-commons.exception :as cx]
+            [clojure-commons.exception-util :as cxu]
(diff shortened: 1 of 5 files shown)


slr71:
No problem at all! Thanks for the review!

slr71:
Thanks! And thanks for the review!

PR #168, opened 2020-07-30
Dashboard schema updates

johnworth:
Adds the content and author fields to the dashboard feed items schema and adds the video feed to the dashboard feeds schema.

Depends on https://github.com/cyverse-de/dashboard-aggregator/pull/1, they should probably be merged at the same time.

### DIFF
--- a/src/terrain/routes/schemas/dashboard_aggregator.clj
+++ b/src/terrain/routes/schemas/dashboard_aggregator.clj
@@ -29,11 +29,14 @@
    (optional-key :parent_id)          (describe (maybe UUID) "The UUID of the parent analysis. Only for batch analyses")})
 
 (defschema DashboardFeedItem
-  {:id          (describe String "The unique identifier for a feed. Probably not a UUID")
-   :name        (describe String "The name of the item in the feed. Not the name of the feed itself")
-   :description (describe String "Corresponds to the content snippet provided by the feed. Named description to match the other types")
-   :link        (describe String "Link to the source of the feed item. Probably leads to the website")
-   :date_added  (describe String "The date the item was published")})
+  {:id                     (describe String "The unique identifier for a feed. Probably not a UUID")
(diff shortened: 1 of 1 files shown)


src/terrain/routes/schemas/dashboard_aggregator.clj
```diff
@@ -29,15 +29,19 @@
...
-   :name        (describe String "The name of the item in the feed. Not the name of the feed itself")
-   :description (describe String "Corresponds to the content snippet provided by the feed. Named description to match the other types")
-   :link        (describe String "Link to the source of the feed item. Probably leads to the website")
-   :date_added  (describe String "The date the item was published")})
+  {:id                     (describe String "The unique identifier for a feed. Probably not a UUID")
+   :name                   (describe String "The name of the item in the feed. Not the name of the feed itself")
+   :description            (describe String "Corresponds to the content snippet provided by the feed. Named description to match the other types")
+   :link                   (describe String "Link to the source of the feed item. Probably leads to the website")
+   :date_added             (describe String "The date the item was added to the feed")
+   :publication_date       (describe String "The date the item was originally published")
```
ianmcorvidae, line 37: Might be nothing, especially since I'm not writing anything consuming this API right now, but just reading this I'm not quite sure how to distinguish these two dates or determine what's user-relevant. Not sure if that's inherent confusion in how the feeds work or if it's something that can be described more clearly.

johnworth: I'm not entirely sure there's a difference, or that the difference matters. I noticed that there were two date fields provided by the rss-parser library, so I decided to expose them both with a description that best matches my understanding of them. It's mostly just covering a case where we might end up needing to differentiate the difference between publication on a site vs. something getting added to the site's feed.

ianmcorvidae: Fair enough, makes sense to include both to me as well

PR #169, opened 2020-08-06
CORE-1045 Add async folder move support to DOI Request endpoints.

psarando:
This PR will add async folder move support to DOI Request endpoints.

Currently the DOI Request endpoints are trying to move the user's folder to the staging or curated directory as the irods proxy user, and then immediately share the moved folder with the DOI curators group using the destination path, which doesn't exist yet since the async task service takes a few seconds to perform the move.

This PR updates these endpoints to spawn a thread that polls the async tasks service until the async move is complete, and then the thread will update the permissions on the moved folder. In the meantime the DOI request or creation will continue as normal while the move happens in the background.

If the move fails for any reason, the permissions will still be updated on the source folder so the DOI curators (or some other admin) can look into the issue and move the folder later.

### DIFF
--- a/src/terrain/clients/async_tasks.clj
+++ b/src/terrain/clients/async_tasks.clj
@@ -1,0 +1,14 @@
+(ns terrain.clients.async-tasks
+  (:require [async-tasks-client.core :as async-tasks-client]
+            [clojure.string :as string]
+            [terrain.util.config :as config]))
+
+(defn run-async-thread
+  [async-task-id thread-function prefix]
+  (let [^Runnable task-thread (fn [] (thread-function))]
+    (.start (Thread. task-thread (str prefix "-" async-task-id))))
+  async-task-id)
+
+(defn get-by-id
+  [id]
+  (async-tasks-client/get-by-id (config/async-tasks-client) id))
--- a/src/terrain/services/permanent_id_requests.clj
+++ b/src/terrain/services/permanent_id_requests.clj
@@ -9,6 +9,7 @@
             [clojure.tools.logging :as log]
             [clojure-commons.file-utils :as ft]
             [org.cyverse.metadata-files.datacite-4-1 :as datacite]
+            [terrain.clients.async-tasks :as async-tasks-client]
             [terrain.clients.data-info :as data-info]
             [terrain.clients.data-info.raw :as data-info-client]
             [terrain.clients.ezid :as ezid]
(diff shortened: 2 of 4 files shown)


src/terrain/clients/async_tasks.clj
```diff
@@ -0,0 +1,14 @@
+(ns terrain.clients.async-tasks
+  (:require [async-tasks-client.core :as async-tasks-client]
+            [clojure.string :as string]
+            [terrain.util.config :as config]))
+
+(defn run-async-thread
+  [async-task-id thread-function prefix]
+  (let [^Runnable task-thread (fn [] (thread-function))]
+    (.start (Thread. task-thread (str prefix "-" async-task-id))))
+  async-task-id)
```
psarando, line 10: This was pretty much just copied out of `data-info`:
https://github.com/cyverse-de/data-info/blob/ebc288d08b3b6fa0df17008d04959bc1d7675535/src/data_info/clients/async_tasks.clj#L38-L42

src/terrain/services/permanent_id_requests.clj
```diff
@@ -184,46 +185,102 @@ If this dataset accompanies a paper, please contact us with the DOI for that pap
...
-
+   (-> (data-info-client/move-single user id dest-path)
+       :body
+       (json/decode true)
+       :async-task-id)
    (catch Object e
      (log/error e)
      (email/send-permanent-id-request-data-move-error path dest-path current-user (:body e (str e)))
-     path)))
+     nil)))
```
psarando, line 195: Now returns the async task ID for the move, rather than the move destination path.

src/terrain/services/permanent_id_requests.clj
```diff
@@ -184,46 +185,102 @@ If this dataset accompanies a paper, please contact us with the DOI for that pap
...
+                                                       id
+                                                       :filter-include "path")))
+          ;; Else keep waiting for up to wait-seconds-max.
+          ;; Since the sleep seconds are doubled each time,
+          ;; then this thread has already been waiting
+          ;; ((sleep-seconds * 2) - 1) number of seconds.
+          (if (< (dec (* sleep-seconds 2)) (config/permanent-id-async-move-wait-seconds-max))
+            (recur (* sleep-seconds 2))
+            (throw+ (str "Async move took too long to complete successfully. Last status: "
+                         last-status))))))
```
psarando, line 223: Hopefully these comments make sense.

src/terrain/util/config.clj
```diff
@@ -413,6 +419,13 @@
   [props config-valid configs]
   "terrain.job-exec.default-output-folder" "analyses")
 
+(cc/defprop-optint permanent-id-async-move-wait-seconds-max
+  "The max amount of time to wait for async folder moves, in seconds.
+   By default, 1023 = (2^10-1) seconds, since the waiting thread doubles its
+   sleep-seconds each time it polls until the async task completes."
+  [props config-valid configs]
+  "terrain.permanent-id.async-move.wait-seconds.max" 1023)  ;; Just over 17 minutes
```
psarando, line 427: Hopefully these comments also make sense.

src/terrain/services/permanent_id_requests.clj
ianmcorvidae: I had to do a bit of mental math to verify the claim that at this point in the loop it'll have slept `(- (* sleep-seconds 2) 1)` seconds, but once I'd satisfied my own curiosity it makes sense. I'm imagining it's not super important that we fit in exactly under the maximum, since this can end up running past it (but only one loop, I think), but mostly just that we don't let it loop forever.

psarando: Yeah, I had to think through the math a couple of times before I got it right. In case it's not obvious to everyone, this come from one of those sum formulas I remember from my CS courses: `∑2^n from 0 to k` is equivalent to `2^(k+1) - 1`.

I couldn't find a decent link describing this sum for these comments, but I did find a very brief description on wikipedia (just the first one in this section): https://en.wikipedia.org/wiki/1_%2B_2_%2B_4_%2B_8_%2B_%E2%8B%AF#Summation

Since the first thing this thread does is the sleep, then at this point in the code `sleep-seconds*2 - 1` = `2^n - 1` if `n` is the number of times the thread has already slept.

Also, I agree that it's not super important that we fit in exactly under the maximum, so the `wait-seconds-max` config could be any value, and you're right we would only pass that max by 1 loop if the config isn't a value equal to `2^n-1`. That could mean this thread could sleep nearly twice as long as the max config (for example if the config were 1024 then thread could end up waiting 1024*2-1=2047 seconds or just over 34 minutes), but that still doesn't seem like that big of a deal for now.

psarando:
> One thing I'm not sure about is, since we're starting an async thread here too, would it be beneficial to have an async-task for this process as well? We don't have a super clear idea of when to do this sort of tracking right now, I think -- clearly, the one for the move is useful since we're using it here, but that particular need is less likely to come up for this case. The vision for async-tasks in general is that we create one for anything we're doing asynchronously, and the service handles keeping tabs on the process, doing anything to restart it if needed, notification, etc. -- but right now it can't do a lot of that stuff. It might be moot, since it wouldn't be too hard to add it later, but I guess I want to plant the seed of the idea at least, heh.

I'm not exactly sure what you mean. Have an async task for the entire DOI request or DOI creation process, email notifications and all?

I didn't think these processes took too long to complete, other than the data move. Though I suppose the submission to the external service for creating the DOI can add a little time, but still I would think that more immediate success/failure feedback is better for these endpoints. Also, the data moves are part of the DOI request/creation SOP, but they're also not critical to the process and can be corrected later by an admin, so I also like the idea that the move can happen in the background while the more critical parts can continue processing synchronously.

ianmcorvidae:
I guess what I basically mean is "should we have an async-task (or more than one) for any stuff that isn't synchronous but which happens due to user request and that we're running" -- in data-info, wherever `run-async-thread` is used, we create a task and wrap it so it sends updates to that task, even if it's something that could be quite short or easily fixed (emptying trash with few items in it, for example). Right now, it's maybe hard to see why that would be valuable, which is why I mentioned the overall vision of having somewhere that the updates, completions, errors, etc. can go that allows us to track or react to those things (or that itself can react to those things). In this case, as you point out, the asynchronous parts aren't super critical and can be corrected later by an admin, but having this tracking could mean those admins have a place to look and see -- did the move itself fail? the permissions after? what error message went through? etc. (all pending implementation, of course).

ianmcorvidae:
(more specifically: in this case I specifically meant creating an async-task for the process that waits for the move to complete and then does the after-move stuff -- that is, everything that runs in the async thread)

psarando:
> (more specifically: in this case I specifically meant creating an async-task for the process that waits for the move to complete and then does the after-move stuff -- that is, everything that runs in the async thread)

OK, then yeah, that makes sense, and it could probably be converted into its own proxy-move-then-fix-perms async task later. I didn't think to make that process its own async task, probably just because I'm not too familiar with that service yet and wouldn't know how to create that task at this point.

psarando:
> ... it wouldn't be too hard to add it later, but I guess I want to plant the seed of the idea at least, heh.

I reviewed (again) what `data-info` is doing to create and update async tasks, and it looks fairly straight forward, so it shouldn't be hard to convert this async thread over to using an async task in the future.

I'll go ahead and merge this for now, since if something goes wrong in the thread, the error is logged and an error email is sent to the DOI curators email list, so admins have somewhere to look for issues. Later this thread could benefit from using async tasks when they can send admins notifications or emails, and maybe then admins can check for task issues in a UI dashboard or something.

psarando:
Thanks for the review! 👍

PR #170, opened 2020-08-31
Bag support

johnworth:
Add routes, a service, a client, and configuration values for bag-related endpoints. The routes are swaggerized.

Fixes the default values for the sessions, preferences, and saved-searches config settings so they hit user-info.

### DIFF
--- a/src/terrain/clients/bags.clj
+++ b/src/terrain/clients/bags.clj
@@ -15,7 +15,7 @@
 
 (defn has-bags
   [username]
-  (<= 200 (:status (http/head (bags-url [username]))) 299))
+  (:body (http/head (bags-url [username]))))
 
 (defn get-bags
   [username]
@@ -30,7 +30,7 @@
 
 (defn delete-all-bags
   [username]
-  (<= 200 (:status (http/delete (bags-url [username]))) 299))
+  (:body (http/delete (bags-url [username]))))
 
 (defn get-bag
   [username id]
@@ -44,4 +44,4 @@
 
 (defn delete-bag
   [username id]
-  (<= 200 (:status (http/delete (bags-url [username id]))) 299))
+  (:body (http/delete (bags-url [username id]))))
--- a/src/terrain/clients/bags.clj
+++ b/src/terrain/clients/bags.clj
@@ -7,11 +7,9 @@
             [terrain.util.config :as config]))
 
 (defn- bags-url
-  ([]
-   (bags-url []))
-  ([components]
-   (-> (apply url (config/bags-base-url) components)
-       (str))))
+[& [components]]
+  (-> (apply url (config/bags-base-url) components)
+      (str)))
 
 (defn has-bags
   [username]
--- a/src/terrain/clients/bags.clj
+++ b/src/terrain/clients/bags.clj
@@ -1,0 +1,47 @@
+(ns terrain.clients.bags
+  (:use [clojure-commons.error-codes]
+        [slingshot.slingshot :only [try+ throw+]])
+  (:require [clj-http.client :as http]
+            [cemerick.url :refer [url]]
+            [clojure.tools.logging :as log]
+            [terrain.util.config :as config]))
+
+(defn- bags-url
+  ([]
+   (bags-url []))
+  ([components]
+   (-> (apply url (config/bags-base-url) components)
+       (str))))
+
+(defn has-bags
+  [username]
+  (<= 200 (:status (http/head (bags-url [username]))) 299))
+
+(defn get-bags
+  [username]
+  (:body (http/get (bags-url [username]) {:as :json})))
+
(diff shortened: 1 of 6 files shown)


src/terrain/clients/bags.clj
```diff
@@ -0,0 +1,47 @@
...
+            [cemerick.url :refer [url]]
+            [clojure.tools.logging :as log]
+            [terrain.util.config :as config]))
+
+(defn- bags-url
+  ([]
+   (bags-url []))
+  ([components]
+   (-> (apply url (config/bags-base-url) components)
+       (str))))
```
slr71, line 14: `apply` will treat `nil` as an empty collection, so you can keep the same functionality without having multiple arities by doing this:

```clojure
(defn- bags-url [& [components]]
 (-> (apply url (config/bags-base-url) components)
     (str))))
```

I tend to like the brevity of the destructuring option, but this is a purely aesthetic choice. Pick whichever option you like best.

src/terrain/clients/bags.clj
```diff
@@ -0,0 +1,47 @@
...
+
+(defn has-bags
+  [username]
+  (:body (http/head (bags-url [username]))))
```
slr71, line 18: In cases where you're not expecting a response body, it's probably better to explicitly return `nil`. That allows readers to quickly determine that the function is being called for side-effects (in this case, either an exception is thrown or not).

src/terrain/routes/bags.clj
```diff
@@ -0,0 +1,59 @@
...
+  (optional-routes
+   [config/bag-routes-enabled]
+
+   (context "/bags" []
+     :tags ["bags"]
+
+     (HEAD "/" []
+       :summary     HasBagsSummary
+       :description HasBagsDescription
+       (ok (has-bags (:username current-user))))
```
slr71, line 22: In cases where you're not returning a response body, it's probably better to move the endpoint body out of the call to the function generating the HTTP response map, for example:

``` clojure
     (HEAD "/" []
       :summary     HasBagsSummary
       :description HasBagsDescription
       (has-bags (:username current-user))
       (ok))
```

Doing this helps to ensure that you don't get any unintended cruft in the response body.

src/terrain/routes/schemas/bags.clj
```diff
@@ -0,0 +1,41 @@
...
+(def BagListDescription "List all of the bags for the user")
+
+(def GetBagSummary "Gets a bag")
+(def GetBagDescription "Gets a single bag for a user by its UUID")
+
+(def DeleteAllBagsSummary "Deletes all of a user's bags")
+(def DeleteAllBagsDescription "Deletes all of a user's bags")
+
+(def DeleteBagSummary "Delete a bag")
+(def DeleteBagDescription "Delete's a bag for a user based on its UUID")
```
slr71, line 41: Tpyo: `Delete's`

src/terrain/util/config.clj
```diff
@@ -489,7 +496,7 @@
 (cc/defprop-optstr prefs-base-url
   "The hostname of the user-preferences service"
   [props config-valid configs]
-  "terrain.preferences.host" "http://user-preferences")
+  "terrain.preferences.host" "http://user-info/preferences")
```
slr71, line 499: 👍

src/terrain/util/config.clj
```diff
@@ -499,12 +506,17 @@
 (cc/defprop-optstr sessions-base-url
   "The hostname of the user-sessions service"
   [props config-valid configs]
-  "terrain.sessions.host" "http://user-sessions")
+  "terrain.sessions.host" "http://user-info/sessions")
```
slr71, line 509: 👍

src/terrain/util/config.clj
```diff
@@ -499,12 +506,17 @@
...
 
 (cc/defprop-optstr saved-searches-base-url
   "The base URL of the saved-searches service"
   [props config-valid configs]
-  "terrain.saved-searches.host" "http://saved-searches")
+  "terrain.saved-searches.host" "http://user-info/searches")
```
slr71, line 514: 👍

src/terrain/routes/schemas/bags.clj
johnworth: This should be fixed now.

src/terrain/routes/bags.clj
johnworth: This should be fixed now.

src/terrain/clients/bags.clj
johnworth: Should be fixed now, thanks.

src/terrain/clients/bags.clj
johnworth: I put this change in, thanks for suggesting it.

PR #171, opened 2020-09-08
Add Default Bag Support

johnworth:
Adds support for default bag handling, which allows callers to ignore the multiple bag support if they so wish.

This is dependent on https://github.com/cyverse-de/user-info/pull/5 getting merge.

### DIFF
--- a/src/terrain/clients/bags.clj
+++ b/src/terrain/clients/bags.clj
@@ -7,7 +7,7 @@
             [terrain.util.config :as config]))
 
 (defn- bags-url
-[& [components]]
+  [& [components]]
   (-> (apply url (config/bags-base-url) components)
       (str)))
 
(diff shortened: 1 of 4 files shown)


src/terrain/clients/bags.clj
```diff
@@ -46,3 +46,19 @@
...
+
+(defn update-default-bag
+  [username contents]
+  (:body (http/post (bags-url [username "default"]) {:form-params  contents
+                                                     :content-type :json
+                                                     :as           :json})))
+
+(defn delete-default-bag
+  [username]
+  (:body (http/delete (bags-url [username "default"])))
```
slr71, line 63: This is a nitpick, but you can probably remove the extraction of the `:body` element from the result of `http/delete` since the value is being discarded anyway.

slr71 (approved):
This looks good to me. I had one minor nitpick, but nothing serious. :shipit: whenever you're ready.

PR #177, opened 2020-10-23
CORE-1009 User can view community data without logging in

ash3rz:
With these changes, in Sonora users should be able to browse community data, read files (textual and tabular), and search community data without logging in.  

Still To Do:
* View details - this involves the stats endpoint which can accept an array of paths and/or an array of UUIDs
* View metadata - requires a UUID
There was talk about converting all the data endpoints over to using paths instead of UUIDs and I didn't want to overlap with that effort.

### DIFF
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -77,6 +77,16 @@
   [^String user ^String dir]
   (cr/ensure-created user dir))
 
+(defn get-request-user
+  "Returns the anonymous user for public data if a user is not provided"
+  [user path]
+  (if-let [request-user (if (string/starts-with? path (cfg/fs-community-data))
+                          (or user "anonymous")
+                          user)]
+    request-user
+    (throw+ {:type :clojure-commons.exception/not-authorized
+             :user user})))
+
 (defn read-chunk
   "Uses the data-info read-chunk endpoint."
   [params body]
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -102,8 +102,10 @@
 (defn manifest
   "Uses the data-info manifest endpoint."
   [params]
-  (let [path-uuid (uuid-for-path (:user params) (:path params))]
-    (raw/manifest (:user params) path-uuid)))
+  (let [path      (:path params)
+        user      (get-request-user (:user params) path)
+        path-uuid (uuid-for-path user path)]
+    (raw/manifest user path-uuid)))
 
 (defn create-dirs
   [{:keys [user]} {:keys [paths]}]
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -90,8 +90,10 @@
 (defn read-chunk
   "Uses the data-info read-chunk endpoint."
   [params body]
-  (let [path-uuid (uuid-for-path (:user params) (:path body))]
-    (raw/read-chunk (:user params) path-uuid (:position body) (:chunk-size body))))
+  (let [path      (:path body)
+        user      (get-request-user (:user params) path)
+        path-uuid (uuid-for-path user path)]
+    (raw/read-chunk user path-uuid (:position body) (:chunk-size body))))
 
 (defn read-tabular-chunk
   "Uses the data-info read-tabular-chunk endpoint."
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -79,13 +79,16 @@
 
 (defn get-request-user
   "Returns the anonymous user for public data if a user is not provided"
-  [user path]
-  (if-let [request-user (if (string/starts-with? path (cfg/fs-community-data))
-                          (or user "anonymous")
-                          user)]
-    request-user
-    (throw+ {:type :clojure-commons.exception/not-authorized
-             :user user})))
+  [user paths]
+  (let [paths (if (= (type paths) String)
+                [paths]
+                paths)
+        has-private-paths? (some #(not (string/starts-with? % (cfg/fs-community-data))) paths)
+        request-user (if has-private-paths?
+                       user
+                       (or user "anonymous"))]
+    (or request-user (throw+ {:type :clojure-commons.exception/not-authorized
+                              :user user}))))
 
 (defn read-chunk
   "Uses the data-info read-chunk endpoint."
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -77,7 +77,7 @@
   [^String user ^String dir]
   (cr/ensure-created user dir))
 
-(defn get-request-user
+(defn get-public-data-user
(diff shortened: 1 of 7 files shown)


src/terrain/clients/data_info.clj
```diff
@@ -77,23 +77,42 @@
...
+(defn get-public-data-user
+  "Returns the anonymous user for public data if a user is not provided"
+  [user paths]
+  (let [paths (if (= (type paths) String)
+                [paths]
+                paths)
+        has-private-paths? (some #(not (string/starts-with? % (cfg/fs-community-data))) paths)
+        request-user (if has-private-paths?
+                       user
+                       (or user "anonymous"))]
```
slr71, line 89: I haven't looked too closely into this yet, but would it be reasonable to let `data-info` validate the permissions on the files instead? The advantage of letting `data-info` do the checks is that it would be possible to use Terrain to access endpoints that are anonymously readable, but outside of the community data collection. The disadvantage is that you'd have to add custom error handling to detect permissions errors and return a 401 in cases where the user wasn't authenticated. It also might take a little bit longer to get 403 responses back if response times are a concern.

src/terrain/clients/search.clj
```diff
@@ -1,13 +1,14 @@
...
-        [terrain.util.transformers :only [secured-params]])
+        [terrain.util.transformers :only [add-current-user-to-map]])
   (:require [clj-http.client :as http]
             [cemerick.url :refer [url]]))
 
 (defn do-data-search
   [body]
-  (let [req-options  {:form-params  body
-                      :query-params (secured-params)
+  (let [query-params {:user (or (:user (add-current-user-to-map {})) "anonymous")}
```
slr71, line 9: I'd be a little tempted to enhance `add-current-user-to-map` to do something like this:

``` clojure
(defn add-current-user-to-map
  "Adds the name and e-mail address of the currently authenticated user to a
   map that can be used to generate a query string. If no user is authenticated,
   the `user` parameter is set to `anonymous` and the rest of the user attribute
   parameters are omitted."
  ([query default-username]
   (->> (assoc query
               :user       (or (:shortUsername current-user) default-username)
               :email      (:email current-user)
               :first-name (:firstName current-user)
               :last-name  (:lastName current-user))
        (remove-vals invalid-query-param-value?)))
  ([query]
   (add-current-user-to-map query "anonymous")))
```

slr71 (approved):
This looks good to me. I had a couple of optional suggestions, but the code seems like it should work fine the way it is. :shipit: whenever you're ready.

src/terrain/clients/data_info.clj
ianmcorvidae: The possible issue here is that a lot of folders are anonymously accessible due to "send to genome browser" (which, IIRC, shares the parent folders too), but folks normally don't expect those to be "browsable". We ran into the same issue with dav-anon and turned off the index pages for it for that reason.

slr71: Ah okay. It probably is best to restrict it to community data then.

src/terrain/clients/data_info.clj
```diff
@@ -77,23 +77,42 @@
   [^String user ^String dir]
   (cr/ensure-created user dir))
 
+(defn get-public-data-user
+  "Returns the anonymous user for public data if a user is not provided"
+  [user paths]
+  (let [paths (if (= (type paths) String)
+                [paths]
+                paths)
```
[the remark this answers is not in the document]
ash3rz: Ah, much more concise!  Thanks!

src/terrain/clients/data_info.clj
ash3rz: That brings up an interesting point though - do those "send to genome browser" folders become searchable for the anonymous user?  If so, then I might need to rethink some things for the search endpoint so only community data results come up.  I did some mild testing and only saw community data results, but I may not be hitting the right search terms in that case.

src/terrain/clients/search.clj
ash3rz: 😄 This is a LOT cleaner - I'll refactor to this. Thanks!

src/terrain/clients/data_info.clj
slr71: They become accessible to the anonymous user, but it sounds like we don't want people who aren't logged in to be able to discover that these files exist. My guess is that we don't want these files to be searchable by the anonymous user.

src/terrain/clients/search.clj
slr71: No problem at all. :)

src/terrain/clients/data_info/raw.clj
```diff
@@ -107,8 +107,8 @@
 
 (defn read-chunk
   "Uses the data-info read-chunk endpoint."
-  [user path-uuid position chunk-size]
-  (request :get ["data" path-uuid "chunks"]
+  [user path position chunk-size]
+  (request :get ["data" "by-path" "chunks" (url/url-encode path)]
```
ash3rz, line 111: Not sure if this is the right thing to do, but I noticed this starting to fail while testing out some refactors (also fails in master) so I switched the endpoint.  I noticed both sonora and the DE are sending paths, so I assumed this would be ok.

ianmcorvidae: That's fine to do since I merged cyverse-de/data-info#44 earlier, but I'm concerned if the UUID versions are failing now. What are you seeing as far as that?

ianmcorvidae: ... I see my mistake, I'll fix that. But this change is still fine, I just don't want to break the old one either, lol

ash3rz: Ah ok, even better - I'm guessing I don't even need this change anymore.  Thanks!  I only tried testing opening a file in sonora and the DE and saw those 2 cases were using paths, but there might be cases where we are using UUIDs instead. 🤔

ianmcorvidae: We'd want to double-check nothing is trying to use a UUID version (I guess it'd have to be using the raw client directly), but ultimately what you're doing here is the plan, in order to avoid the UUID overhead -- the "uuid-for-path" you're doing here is another data-info call, and then if a UUID is passed to data-info it has to turn it back into a path anyway. So I think you're good either way.

src/terrain/clients/data_info.clj
ash3rz: To close the loop on this one, we decided this should be added to the `search` service so that it can better handle modifying the query to narrow it down to only community data and also handle contradictory clauses and whatnot.  A ticket was created for that work.

src/terrain/clients/data_info/raw.clj
ash3rz: Ah ok, I removed this change for this PR since the old endpoint started working and I didn't do _that_ much research into what might still be using the UUID version 😄

ash3rz:
Thanks for the reviews!

PR #178, opened 2020-10-27
CORE-1143: added a way to return a list of apps the user has recently used.

slr71:
This change adds `recentlyUsed` to the response body schema for the dashboard endpoint. It also adds a new optional query parameter, `start-date-interval`, to the same endpoint.

### DIFF
--- a/src/terrain/routes/schemas/dashboard_aggregator.clj
+++ b/src/terrain/routes/schemas/dashboard_aggregator.clj
@@ -4,60 +4,139 @@
   (:import [java.util UUID]))
 
 (defschema DashboardApp
-  {:id                              (describe String "The app ID")
-   :system_id                       (describe String "The system ID for the app")
-   (optional-key :name)             (describe (maybe String) "The name of the app")
-   (optional-key :description)      (describe (maybe String) "The description of the app")
-   :username                        (describe String "The name of the user that created the app")
-   (optional-key :wiki_url)         (describe (maybe String) "The URL to the wiki entry for the app")
-   (optional-key :integration_date) (describe (maybe String) "The date the app was integrated. Milliseconds since epoch")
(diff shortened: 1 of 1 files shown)


src/terrain/routes/schemas/dashboard_aggregator.clj
```diff
@@ -4,60 +4,142 @@
...
+   (describe (maybe [DashboardFeedItem]) "The videos feed")})
 
 (defschema DashboardAggregatedApps
-  {(optional-key :recentlyAdded) (describe (maybe [DashboardApp]) "Apps recently added by the user")
-   :public                       (describe [DashboardApp] "Apps recently made public")})
+  {(optional-key :recentlyAdded)
+   (describe (maybe [DashboardApp]) "Apps recently added by the user")
+
+   (optional-key :recentlyRan)
+   (describe (maybe [DashboardApp]) "Public apps that were recently used in jobs")
```
slr71, line 121: This is the only substantive change in this PR.

ash3rz: I probably missed it in the other PR since I'm not super familiar with app queries, but will this only return recently run public apps and not private apps as well?  This might go back to your comment in the other PR about whether this feature is supposed to show trending apps or recently used apps (which I'm not 100% sure, but thought it was for recently used apps).

sri2k1us: The feature I requested was recently used apps by user (public or private). Trending apps will need to come from the `analytics`

slr71: Yeah, I implemented the wrong feature. 😖

sri2k1us (approved):
👍 LGTM.

slr71:
Thanks for the review!

PR #179, opened 2020-11-17
CORE-1173: switched to a more recent version of clj-http

slr71:
This fixes a `stream closed` error that's intermittently occurring when we make calls to endpoints that return bare JSON arrays (that is, JSON arrays that aren't wrapped in an object).


slr71:
I'm going to merge this immediately, since it's just a dependency update.

PR #180, opened 2020-11-20
Remove unauthd access to data roots endpoint

ash3rz:
Since sonora will no longer need to hit the data roots endpoint unauthenticated, I removed the logic for it.


slr71 (approved):
LGTM! :shipit:

ash3rz:
Thanks for the review!

PR #182, opened 2020-11-23
CORE-1009 Allow logged out users to use /filesystem/stat endpoint

ash3rz:
This should be the last change for CORE-1009 - allowing logged out users to browse public data

### DIFF
--- a/src/terrain/clients/data_info.clj
+++ b/src/terrain/clients/data_info.clj
@@ -77,22 +77,11 @@
   [^String user ^String dir]
   (cr/ensure-created user dir))
 
-(defn get-public-data-user
-  "Returns the anonymous user for public data if a user is not provided"
-  [user paths]
-  (let [paths (if (sequential? paths) paths [paths])
-        has-private-paths? (some #(not (string/starts-with? % (cfg/fs-community-data))) paths)
-        request-user (if has-private-paths?
-                       user
-                       (or user "anonymous"))]
-    (or request-user (throw+ {:type :clojure-commons.exception/not-authorized
-                              :user user}))))
-
 (defn read-chunk
   "Uses the data-info read-chunk endpoint."
   [params body]
   (let [path      (:path body)
-        user      (get-public-data-user (:user params) path)
+        user      (st/get-public-data-user (:user params) path)
         path-uuid (uuid-for-path user path)]
     (raw/read-chunk user path-uuid (:position body) (:chunk-size body))))
 
@@ -100,7 +89,7 @@
   "Uses the data-info read-tabular-chunk endpoint."
   [params body]
   (let [path      (:path body)
-        user      (get-public-data-user (:user params) path)
+        user      (st/get-public-data-user (:user params) path)
         path-uuid (uuid-for-path user path)]
     (raw/read-tabular-chunk user path-uuid (:separator body) (:page body) (:chunk-size body))))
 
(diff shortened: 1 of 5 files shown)


src/terrain/clients/data_info.clj
```diff
@@ -77,38 +77,27 @@
   [^String user ^String dir]
   (cr/ensure-created user dir))
 
-(defn get-public-data-user
```
ash3rz, line 80: Most of the changes are moving this function - keeping this here, ended up with a circular reference, so I moved it

src/terrain/services/filesystem/stat.clj
```diff
@@ -115,6 +128,9 @@
 
 (defn do-stat
   [{user :user} body]
-  (-> (data-raw/collect-stats user body)
-      :body
-      (json/decode true)))
+  (let [paths         (:paths body)
+        ids           (:ids body)
+        request-user  (if ids user (get-public-data-user user paths))]
```
ianmcorvidae, line 133: This is a little finicky, and I don't really mind if it stays as it currently is, but it might be worth adding an extra arity to `get-public-data-user` -- my thinking is, support `[user paths]` like you have, but also `[user paths ids]`, where if `ids` is empty, use the `[user paths]` code, otherwise throw the `not-authorized` error like you are. That way, there's not two different pathways for errors -- I think right now, if you provide private paths, it'll throw not authorized above, but if you pass IDs it will probably throw an error for the missing "user" query parameter in data-info? Maybe? We're transitioning away from using IDs, but I think it's still nicer to be consistent.

Probably to do this, I'd put the bulk of the code into the `[user paths ids]` arity, and have `[user paths]` arity just be `([user paths] (get-public-data-user user paths nil))`

In case "multi-arity function" is meaningless to you, or you need a refresher, since I know you write less clojure than some: http://clojure-doc.org/articles/language/functions.html#multi-arity-functions

src/terrain/services/filesystem/stat.clj
```diff
@@ -67,6 +69,17 @@
...
+(defn get-public-data-user
+  "Returns the anonymous user for public data if a user is not provided"
+  [user paths]
+  (let [paths (if (sequential? paths) paths [paths])
+        has-private-paths? (some #(not (string/starts-with? % (cfg/fs-community-data))) paths)
+        request-user (if has-private-paths?
+                       user
+                       (or user "anonymous"))]
+    (or request-user (throw+ {:type :clojure-commons.exception/not-authorized
+                              :user user}))))
```
ianmcorvidae, line 81: Just to make the other comment I made easier, here's what I'd do here:

```suggestion
(defn get-public-data-user
  "Returns the anonymous user for public data if a user is not provided"
  ([user paths ids]
   (let [paths (if (sequential? paths) paths [paths])
         has-private-paths? (or (seq ids) (some #(not (string/starts-with? % (cfg/fs-community-data))) paths))
         request-user (if has-private-paths?
                        user
                        (or user "anonymous"))]
     (or request-user (throw+ {:type :clojure-commons.exception/not-authorized
                               :user user}))))
  ([user paths]
   (get-public-data-user user paths nil))
```

untested code so make sure I didn't make a typo, but anyway :smile:

ianmcorvidae (approved):
Fine as is, also had one thought if it appeals to you.

src/terrain/services/filesystem/stat.clj
ash3rz: Cool - thanks for this!  I'll refactor 👍 
In Clojure, I still don't quite have a concept of what's finicky or not, so comments like this are helpful for me 😄

ianmcorvidae (approved):
Just to verify, new changes look good to me too.

psarando (approved):
LGTM 👍

ash3rz:
Thanks for the reviews!

PR #185, opened 2020-12-02
CORE-1180: add unit tests for an endpoint that makes calls to another service.

slr71:
Integration tests can be painful to set up and tear down, and it's frequently beneficial to have a tighter feedback loop when code is being developed. This change defines a couple of test fixtures to make it easier to define unit tests involving code that makes calls to other services. It also provides an example unit test for future reference.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -35,7 +35,7 @@
                  [org.cyverse/async-tasks-client "0.0.3"]
                  [org.cyverse/clj-icat-direct "2.9.0"]
                  [org.cyverse/clj-jargon "2.8.9"]
-                 [org.cyverse/clojure-commons "3.0.5"]
+                 [org.cyverse/clojure-commons "3.0.6-SNAPSHOT"]
                  [org.cyverse/cyverse-groups-client "0.1.8"]
                  [org.cyverse/common-cfg "2.8.1"]
                  [org.cyverse/common-cli "2.8.1"]
@@ -53,7 +53,8 @@
             [swank-clojure "1.4.2" :exclusions [org.clojure/clojure]]
             [test2junit "1.2.2"]
             [jonase/eastwood "0.3.10"]]
-  :profiles {:dev     {:resource-paths ["conf/test"]}
+  :profiles {:dev     {:dependencies [[clj-http-fake "1.0.3"]]
+                       :resource-paths ["conf/test" "test-resources"]}
              :uberjar {:aot :all}}
   :main ^:skip-aot terrain.core
   :ring {:handler terrain.core/dev-handler
--- a/project.clj
+++ b/project.clj
@@ -35,7 +35,7 @@
                  [org.cyverse/async-tasks-client "0.0.3"]
                  [org.cyverse/clj-icat-direct "2.9.0"]
                  [org.cyverse/clj-jargon "2.8.9"]
-                 [org.cyverse/clojure-commons "3.0.6-SNAPSHOT"]
+                 [org.cyverse/clojure-commons "3.0.6"]
                  [org.cyverse/cyverse-groups-client "0.1.8"]
                  [org.cyverse/common-cfg "2.8.1"]
                  [org.cyverse/common-cli "2.8.1"]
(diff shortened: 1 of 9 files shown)


project.clj
```diff
@@ -35,7 +35,7 @@
                  [org.cyverse/async-tasks-client "0.0.3"]
                  [org.cyverse/clj-icat-direct "2.9.0"]
                  [org.cyverse/clj-jargon "2.8.9"]
-                 [org.cyverse/clojure-commons "3.0.5"]
+                 [org.cyverse/clojure-commons "3.0.6-SNAPSHOT"]
```
slr71, line 38: I'll change this when the new release of `org.cyverse/clojure-commons` is published.

project.clj
```diff
@@ -53,7 +53,8 @@
             [swank-clojure "1.4.2" :exclusions [org.clojure/clojure]]
             [test2junit "1.2.2"]
             [jonase/eastwood "0.3.10"]]
-  :profiles {:dev     {:resource-paths ["conf/test"]}
+  :profiles {:dev     {:dependencies [[clj-http-fake "1.0.3"]]
+                       :resource-paths ["conf/test" "test-resources"]}
```
slr71, line 57: The `:dev` profile is used both in the REPL and when running `lein test`.

src/terrain/clients/notifications.clj
```diff
@@ -15,11 +15,11 @@
...
                            (add-current-user-to-map query) relative-url)))
 
-(def ^:private last-ten-messages-params
+(def last-ten-messages-params
   {:limit      10
    :offset     0
-   :sort-field :timestamp
-   :sort-dir   :desc})
+   :sort-field "timestamp"
+   :sort-dir   "desc"})
```
slr71, line 22: Using keywords in the query parameter values map wasn't working with `clj-http-fake`.

src/terrain/util/config.clj
```diff
@@ -174,7 +174,7 @@
 (cc/defprop-optstr notificationagent-base-url
   "The base URL to use when connecting to the notification agent."
   [props config-valid configs notification-routes-enabled]
-  "terrain.notificationagent.base-url" "http://notification-agent")
+  "terrain.notificationagent.base-url" "http://notifications")
```
slr71, line 177: This is the base URL used by the new notifications service in Kubernetes.

test/terrain/notifications_test.clj
```diff
@@ -0,0 +1,32 @@
...
+
+(use-fixtures :once test-fixtures/with-test-config test-fixtures/with-test-user)
+
+(defn- notification-url [& components]
+  (str (apply curl/url (config/notificationagent-base-url) components)))
+
+(deftest last-ten-messages
+  (with-fake-routes-in-isolation
+    {{:address      (notification-url "messages")
+      :query-params (map-vals str (add-current-user-to-map nc/last-ten-messages-params))}
```
slr71, line 20: `clj-http-fake` doesn't match query parameters with values that aren't strings, so we coerce the values in the map to be strings. I considered creating a custom function for this, but `map-vals` does exactly what we need relatively simply. Instead of adding a custom function, I'm planning to add a blurb to the contribution guidelines.

src/terrain/util/config.clj
ianmcorvidae: Would it be worth changing the property and `defprop` name here as well, presumably to `terrain.notifications.base-url` and `notifications-base-url`? That would be more consistent with our other services, so it's one fewer "oh, right, it's named that because of <ancient history>" thing to deal with in the future.

If it's out of scope for this PR though, that's fine, just came to mind.

ianmcorvidae (approved):
One thing unrelated to this PR, so this looks good to me.

src/terrain/util/config.clj
slr71: That's a good idea. I'll update them as well.

slr71:
Thanks for the review!

psarando (commented):
Very nice 👍 🎉

slr71:
Thanks for the review!

PR #186, opened 2020-12-03
CORE-1158 Allow public apps to be browsable by logged out users

ash3rz:
Since we're letting Sonora handle creating the categories and showing a 401 for anything other than the `Browse All Apps` category, I didn't need to open up those endpoints.


src/terrain/routes/metadata.clj
```diff
@@ -17,6 +17,7 @@
         [common-swagger-api.schema.apps.rating]
         [common-swagger-api.schema.integration-data :only [IntegrationData]]
         [ring.util.http-response :only [ok]]
+        [terrain.auth.user-attributes]
```
ianmcorvidae, line 20: Small, but unless there's a reason you need to include the whole namespace, this should probably be:

```suggestion
        [terrain.auth.user-attributes :only [require-authentication]]
```

Most clojure lint tools complain about "unlimited `use`" since without being explicit about it, it's relatively easy to end up with weird stuff in your namespace, and it's more opaque.

src/terrain/routes/analyses.clj
```diff
@@ -14,6 +14,7 @@
                 UpdateQuickLaunchGlobalDefault
                 NewQuickLaunchGlobalDefault]]
         [ring.util.http-response :only [ok]]
+        [terrain.auth.user-attributes]
```
ianmcorvidae, line 17: (same comment here, of course)

```suggestion
        [terrain.auth.user-attributes :only [require-authentication]]
```

ianmcorvidae (approved):
One clojure nitpick, otherwise all looks fine to me.

src/terrain/routes.clj
```diff
@@ -75,8 +75,10 @@
    the next set of routes (`secured-routes-no-context`) if nothing matches."
   []
   (util/flagged-routes
+   (apps-routes)
    (dashboard-aggregator-routes)
    (filesystem-stat-routes)
+   (quicklaunch-routes)
```
psarando, line 81: Should these routes now be removed from `secured-routes-no-context` below?

ash3rz: Ah, I didn't notice that section... I guess I don't know what it means to remove it from that section.  I tried removing them locally and nothing seemed to error out, endpoints still seem to work, so I'm not immediately seeing an issue with removing them from `secured-routes-no-context`.

psarando: So all the routes have to be defined in `swagger-routes` at the bottom of this file, which is just using the route wrapper `terrain-routes`.

If you check out `terrain-routes`, you'll see the order the routes are actually defined in.
Since `optionally-authenticated-routes` are listed before `secured-routes-no-context`, then compojure-api will find these routes included in `optionally-authenticated-routes` first.

So it's not an error to also include the duplicate routes in `secured-routes-no-context`, but I think it's cruft that can be removed.

ash3rz: Cool - thanks for the explanation!  I'm guessing it's only really a duplicate if the route originally appeared in `secured-routes-no-context` and not `secured-routes`?  Otherwise, I have more to dupes clean up 😄

psarando: Ah, yes, good catch. Adding the routes to `secured-routes` does add the extra `/secured` context, so that does make a difference and the duplicates could be left in there.

ash3rz:
Thanks for the reviews!

PR #189, opened 2020-12-07
Revert do-stat to use add-current-user-to-map with no default user

ash3rz:
The /filesystem/directory/create endpoint also uses this logic so the user param needs to be consistent


psarando (approved):
LGTM :shipit:

ash3rz:
Thanks for the review!

PR #190, opened 2020-12-09
Rearrange app-ontology-routes prior to apps-routes

ash3rz:
Without this, it's possible for the /apps/hierarchies/:root-iri calls to match /apps/:system-id/:app-id instead

### DIFF
--- a/src/terrain/routes.clj
+++ b/src/terrain/routes.clj
@@ -75,6 +75,7 @@
    the next set of routes (`secured-routes-no-context`) if nothing matches."
   []
   (util/flagged-routes
+   (app-ontology-routes)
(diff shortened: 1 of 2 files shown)


ianmcorvidae (approved):
I think this looks good. I tried to poke around a little and see if I could find anything else that'd get shadowed by `/apps/:system-id/:app-id` but everything else I could find was either fewer or more path components and thus wouldn't get hit by it. I think you already said you were poking around to see if anything else broke, so I'm satisfied if you are!

ash3rz:
The QA tests have taken a major hit, so we should hopefully have a complete picture either way 😄 
Thanks for the help and for the review!

PR #191, opened 2020-12-09
Preserve ordering of all app routes

ash3rz:
This should be the same order as before where all routes that came before `apps-routes` are now correctly before `apps-routes` again.  This is to workaround an issue where some fixed routes (e.g. /apps/hierarchies/:root-iri) are being caught by more dynamic routes (e.g. /apps/:system-id/:app-id)

### DIFF
--- a/src/terrain/routes.clj
+++ b/src/terrain/routes.clj
@@ -75,14 +75,21 @@
    the next set of routes (`secured-routes-no-context`) if nothing matches."
   []
   (util/flagged-routes
-   (app-ontology-routes)
-   (apps-routes)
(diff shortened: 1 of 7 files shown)


ianmcorvidae (approved):
I haven't checked in super detail, but the basic rearrangement looks good and if you've added `require-authentication` to everything (which I'm sure you've checked) that'll be good.

ash3rz:
I checked twice so hopefully I didn't miss anything 😄 
Thanks for the review!

psarando (approved):
Yep, looks right to me 👍

PR #194, opened 2021-01-14
Add rudimentary opentelemetry instrumentation

ianmcorvidae:
This means that if terrain is set up with a proper tracer (usually, using the java autoinstrumentation), it'll create spans and send the trace parent information when it makes HTTP calls, as well as parse trace parents out of HTTP headers. I also added some specific spans for authentication-related code, mostly because when I was testing this I was curious where some time was going.


slr71 (approved):
Looks good to me! :shipit: whenever you're ready.

ianmcorvidae:
Cool, thanks for the lookover. I'll merge it now.

PR #198, opened 2021-02-04
CORE-1226: a bug fix and a miscellaneous testing improvement.

slr71:
These changes aren't related directly to [CORE-1226](https://cyverse.atlassian.net/browse/CORE-1226,) but I did notice the bug and use the improvement to `terrain.auth.user-attributes/with-user` when I was researching the issue.


ianmcorvidae (approved):
LGTM.

slr71:
Thanks for the review!

PR #200, opened 2021-02-24
Add support for instant launch metadata endpoints

johnworth:
Adds support for instant launch metadata endpoints.

### DIFF
--- a/src/terrain/clients/app_exposer.clj
+++ b/src/terrain/clients/app_exposer.clj
@@ -155,9 +155,11 @@
       (client/delete {:as :json})
       (:body)))
 
-;;; We bypass the query map handling built into (app-exposer-url) to use the
-;;; one from clj-http because the former does not handle having an seq of
-;;; values for an entry while the latter does.
+;;; For the (list-metadata) function, we bypass the query map handling built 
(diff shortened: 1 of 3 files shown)


ianmcorvidae (approved):
Seems straightforward to me. Haven't looked at app-exposer, but probably safe to assume the paths and stuff are correct.

src/terrain/clients/app_exposer.clj
```diff
@@ -153,4 +153,36 @@
   [id]
   (-> (app-exposer-url ["instantlaunches" id] {} :no-user true)
       (client/delete {:as :json})
+      (:body)))
+
+;;; We bypass the query map handling built into (app-exposer-url) to use the
+;;; one from clj-http because the former does not handle having an seq of
+;;; values for an entry while the latter does.
+(defn list-metadata
```
ianmcorvidae, line 161: Oh, I guess the one question would be if this should specify that it's for instant launch metadata specifically. Neglected to notice it's not in an IL-specific namespace

johnworth: I added a bit of clarification to the comment.

PR #206, opened 2021-03-11
Add instantLaunches to the dashboard listing

johnworth:
Adds instant launches to the dashboard listing endpoint, but only for logged in users.


slr71 (approved):
Looks good to me! You might want to do a rebase before merging the code, but this should be ready to go at any time. :shipit:

PR #207, opened 2021-03-12
CORE-1339: added the `created_date`, `status`, and `updated_date` fie…

slr71:
…lds to the response bodies for request listings and submissions


slr71:
These are just schema updates, and they shouldn't affect any other services. I'm going to merge this one immediately. Reviews are still welcome.

PR #208, opened 2021-03-12
CORE-1340: changed the request status code name, `complete`, to `appr…

slr71:
…oved`


slr71:
This change is tiny, so I'm going to merge it immediately. Reviews are still welcome.

PR #209, opened 2021-03-19
CORE-1344: added a new endpoint specifically for approving VICE acces…

slr71:
…s requests that allows the number of concurrent jobs to be overridden

See [the Jira issue](https://cyverse.atlassian.net/browse/CORE-1344) for more details.


johnworth (approved):
:shipit:

slr71:
Thanks for the review!

PR #213, opened 2021-04-23
CORE-1372: added the ability to reject a VICE access request after it…

slr71:
… has already been approved

### DIFF
--- a/src/terrain/services/requests.clj
+++ b/src/terrain/services/requests.clj
@@ -67,13 +67,37 @@
   "Marks a request as being in progress."
   (request-update-fn "Your request is in progress." "in-progress"))
 
-(def request-rejected
+(def mark-request-rejected
   "Marks a request as having been rejected."
   (request-update-fn "No deinal reason given." "rejected"))
 
 (def mark-request-approved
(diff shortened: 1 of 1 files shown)


src/terrain/services/requests.clj
```diff
@@ -67,14 +67,38 @@
   "Marks a request as being in progress."
   (request-update-fn "Your request is in progress." "in-progress"))
 
-(def request-rejected
+(def mark-request-rejected
   "Marks a request as having been rejected."
   (request-update-fn "No deinal reason given." "rejected"))
```
ash3rz, line 72: Looks like a tpyo here (outside of your PR changes, but figured I'd mention it):
```suggestion
  (request-update-fn "No denial reason given." "rejected"))
```

ash3rz (approved):
LGTM 👍

src/terrain/services/requests.clj
slr71: Ah, I completely missed that. Thanks for catching it.

slr71:
Thanks for the review!

PR #215, opened 2021-05-13
CORE-1405 Add optional intercom_profile key to ViceRequestDetails

ash3rz:


psarando (approved):
LGTM 👍

ash3rz:
Thanks for the review!
It seems we don't need this change anymore, so I'm going to close this.

PR #222, opened 2021-09-21
CORE-1482 Add Featured Apps endpoint

ash3rz:
Also added the `default-username` as a possible email for cases e.g. apps where you can have anonymous access but it requires an email to also be provided.


psarando (approved):
LGTM 👍

PR #223, opened 2021-09-30
CORE-1484 Add non-admin instant launch metadata search endpoint

ash3rz:
This endpoint is needed for [CORE-1484](https://cyverse.atlassian.net/browse/CORE-1484). It's a duplicate of the admin version of the endpoint except this one is for non-admins and is accessible to users who aren't logged in. This will allow the DE to query for the list of instant launches which have been added to the navigation drawer as a "first class" resource, specifically the `cli` instant launch.

### DIFF
--- a/src/terrain/routes/instantlaunches.clj
+++ b/src/terrain/routes/instantlaunches.clj
@@ -42,13 +42,12 @@
        :return FullInstantLaunchList
        (ok (get-full-instant-launch-list)))
 
-     (context "/metadata" []
-       (GET "/full" []
-         :summary ListFullMetadataSummary
-         :description ListFullMetadataDescription
-         :query [query MetadataListingQueryMap]
-         :return FullInstantLaunchList
-         (ok (list-full-metadata query (or (:username current-user) "anonymous")))))
+     (GET "/metadata/full" []
+       :summary ListFullMetadataSummary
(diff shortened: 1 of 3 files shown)


src/terrain/routes/instantlaunches.clj
```diff
@@ -22,26 +22,38 @@
...
        :return FullInstantLaunchList
        (ok (get-full-instant-launch-list)))
 
+     (context "/metadata" []
+       (GET "/full" []
+         :summary ListFullMetadataSummary
+         :description ListFullMetadataDescription
+         :query [query MetadataListingQueryMap]
+         :return FullInstantLaunchList
+         (ok (list-full-metadata query (or (:username current-user) "anonymous")))))
```
psarando, line 51: I don't think there needs to be another `context` here, since it's only used by 1 endpoint:
```suggestion
       (GET "/metadata/full" []
         :summary ListFullMetadataSummary
         :description ListFullMetadataDescription
         :query [query MetadataListingQueryMap]
         :return FullInstantLaunchList
         (ok (list-full-metadata query (or (:username current-user) "anonymous"))))
```

psarando (approved):
LGTM 👍 

I just had a minor code-style suggestion.

ash3rz:
Thanks for the review!

PR #225, opened 2021-09-30
CORE-1542: updated the location of the signing keys for JWTs

slr71:
This PR makes the following changes:

1. Removes the explicit item names from the volume for the `signing-keys` secret.
2. Changes the mount point for the `signing-keys` secret.
3. Adds a volume and mount point for the new `accepted-keys` secret.
4. Updates the authentication code to log a message every time a legacy JWT token is used for authentication. This will allow us to determine whether or not this feature is being used and possibly remove support altogether at some point in the future.

The goal of this change is to make external deployments easier. External DE deployments currently must have custom deployment definitions for Terrain because they don't have the same set of accepted JWT signing keys. This change removes this requirement. Please see [CORE-1542][1] for details.

[1]: https://cyverse.atlassian.net/browse/CORE-1542

[CORE-1542]: https://cyverse.atlassian.net/browse/CORE-1542

### DIFF
--- a/k8s/terrain.yml
+++ b/k8s/terrain.yml
@@ -36,38 +36,9 @@ spec:
         - name: signing-keys
           secret:
             secretName: signing-keys
-            items:
-              - key: private-key.pem
-                path: private-key.pem
-                mode: 0644
-
-              - key: public-key.pem
-                path: public-key.pem
-                mode: 0644
-
-              - key: QApubkey.pem
-                path: accepted_keys/QApubkey.pem
-                mode: 0644
-
-              - key: apim_staging.pub
-                path: accepted_keys/apim_staging.pub
-                mode: 0644
-
-              - key: legfed-pubkey.pem
-                path: accepted_keys/legfed-pubkey.pem
-                mode: 0644
-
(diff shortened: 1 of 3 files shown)


k8s/terrain.yml
```diff
@@ -36,38 +36,9 @@ spec:
         - name: signing-keys
           secret:
             secretName: signing-keys
-            items:
-              - key: private-key.pem
-                path: private-key.pem
-                mode: 0644
```
ash3rz, line 42: It looks like we might be losing these explicit permission settings on each file - will that be an issue if so?

ash3rz (approved):
I had a question, but not a blocker in any way.  LGTM :shipit:

k8s/terrain.yml
slr71: That shouldn't be an issue. The file secret was being mounted in read-only mode before, so the write permissions weren't relevant anyway. Terrain also runs as root inside the container, so it should be able to read all of the files without any problem.

slr71:
Thanks for the review!

PR #229, opened 2021-10-14
CORE-1488 Fix permanent-id-requests listing endpoint sort params

psarando:
Enum param keys need to be converted to strings before forwarding to the metadata service, or else they will be encoded like `http://metadata/permanent-id-requests?sort-field=:date_submitted`


psarando:
This is a fairly minor fix, so I'll go ahead and merge it right away, but reviews are still welcome 🙇

PR #230, opened 2021-10-16
CORE-1554 Send HTML DOI Completion email to requesting user

psarando:
This PR will update the `permanent-id-requests` completion endpoint to send an HTML DOI Completion email to the requesting user, using the new template added by cyverse-de/de-mailer#10.

This PR will also add the following fixes and updates:
* Rename the permanent-id-request completion email template key `identifiers` to `api_response`, since it's now a JSON response from the DataCite API and not just `DOI` and `ARK` IDs.
* Fix for including the name of the requesting user in data-move error emails.
* Rename the email template keys `username` to `user`, since that's what the de-mailer service is now using in its templates, but it also makes more sense since the value is the user's full name, and not their account username.

### DIFF
--- a/src/terrain/util/email.clj
+++ b/src/terrain/util/email.clj
@@ -79,6 +79,20 @@
       "permanent_id_request_complete"
       template-values)))
 
+(defn send-permanent-id-request-complete-for-user
+  "Sends an email message informing the requesting user of a Permanent ID Request completion."
+  [request-type doi path {:keys [name email]}]
+  (let [template-values {:user name
+                         :path path
+                         :doi  doi}
+        subject         (str request-type " Permanent ID Request Complete")]
+    (send-email
+      :to        email
+      :from-addr (config/permanent-id-request-src-addr)
+      :subject   subject
+      :template  "permanent_id_request_completion_user"
+      :values    template-values)))
+
 (defn send-permanent-id-request-submitted
   "Sends an email message to the user requesting a new Permanent ID Request."
   [request-type path {:keys [commonName email]}]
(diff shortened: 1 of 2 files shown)


psarando (commented):
The new completion email template added by cyverse-de/de-mailer#10 is based on the notification message that is still sent to the requesting user in the DE on completion of their DOI Request:
https://github.com/cyverse-de/terrain/blob/08329d4fb21710352726da0ab58759fec44bb3a7/src/terrain/services/permanent_id_requests.clj#L27-L35

So now they will be getting the same email message twice, a plain-text version in the notification email, and a new HTML email. I think we should still include at least some of this info in the completion notification, though I'm not sure how to modify the message at the moment.

The notification body is not currently viewable within the DE UI, and the user can only see it in the notification email, but this message is viewable in the DOI admin view, by viewing the request's status history. If we ever expose a DOI request listing for users (which is supported in the API https://qa.cyverse.org/terrain/docs/index.html#!/permanent45id45requests/get_terrain_permanent_id_requests), then they will be able to view their request histories and see this message there as well.

For example, this is a screenshot of the DE notification email (displayed in GMail):
Notification email

---

And the new HTML completion email:
HTML email

src/terrain/util/email.clj
```diff
@@ -56,33 +57,48 @@
...
+      :from-addr (config/permanent-id-request-src-addr)
+      :subject   subject
+      :template  "permanent_id_request_completion_user"
+      :values    template-values)))
+
 (defn send-permanent-id-request-submitted
   "Sends an email message to the user requesting a new Permanent ID Request."
   [request-type path {:keys [commonName email]}]
-  (let [template-values {:username     commonName
+  (let [template-values {:user         commonName
```
psarando, line 101: This function sends an email to the user submitting the new request, so the email does not need to display the user's own `username`. Additionally, the new HTML `header` template starts messages by addressing the user with their name pulled from the `user` template value.

PR #231, opened 2021-11-03
Resource Usage Endpoints

johnworth:
Adds admin and non-admin endpoints for getting resource-usage information and posting updates to the CPU hours totals.

Right now hitting the updated totals endpoints doesn't return anything other than a 200 status on a successful update since the actual process for updating the totals is asynchronous. This may change in the future, but for now this should be sufficient for a first pass to get the feature working.

Editing of events and workers isn't supported, but may get added in the future. For now, if you need to undo a CPU hour change, just add another event that undoes it. That's actually the preferred way of doing that kind of update, anyway.


src/terrain/clients/resource_usage_api.clj
```diff
@@ -0,0 +1,103 @@
...
+       (assoc :query query)
+       str)))
+
+(defn current-cpu-hours-total
+  ([]
+   (-> (resource-usage-api ["admin" "cpu" "totals"])
+       (http/get {:as :json})
+       (:body)))
+  ([username]
+   (-> (resource-usage-api [username "cpu" "total"])
```
ianmcorvidae, line 20: I don't know if this can really come up, but I guess if `"admin"` is passed into this function as `username`, it'll behave like the admin version. I guess it might be worth excluding that case so the admin version is definitely only the 0-argument version (I figure just wrapping this side of the function in a `(when-not (= username "admin") )` would be sufficient, we don't need it to actually return a real value)

ianmcorvidae: (this would apply to anything that has the same structure too, I guess)

ianmcorvidae (approved):
This all looks reasonable to me, just my one comment about the client functions for you to consider but I think this looks fine

src/terrain/clients/resource_usage_api.clj
johnworth: Ooh, good point. I'll get a fix in. Thanks!

johnworth:
Thanks for the review!

src/terrain/routes/schemas/resource_usage_api.clj
```diff
@@ -0,0 +1,117 @@
...
+                      [:name
+                       :active
+                       :activation_expires_on
+                       :deactivated_on
+                       :getting_work
+                       :getting_work_on
+                       :getting_work_expires_on
+                       :working
+                       :working_on])]
+          [(optional-key (key keyval)) (val keyval)])))
```
psarando, line 73: I think this could be handled by `schema-tools.core/optional-keys`.
For example:
https://github.com/cyverse-de/terrain/blob/09097dc798d0cbe519ba65e511c97dfb20a6d0c8/src/terrain/routes/schemas/filesystem/tickets.clj#L11-L14

src/terrain/routes/schemas/resource_usage_api.clj
```diff
@@ -0,0 +1,117 @@
...
+                                         :created_by
+                                         :claimed
+                                         :claimed_by
+                                         :claim_expires_on
+                                         :claimed_on
+                                         :processed
+                                         :processing
+                                         :max_processing_attempts
+                                         :attempts])]
+          [(optional-key (key keyval)) (val keyval)])))
```
psarando, line 117: I think this also could be handled by `schema-tools.core/optional-keys`.

psarando (commented):
LGTM 👍

src/terrain/routes/schemas/resource_usage_api.clj
johnworth: Oops, missed that. Thanks!

PR #232, opened 2021-11-11
CORE-1603 Allow null folder values in DOI Request listing responses

psarando:
This PR will update the `PermanentIDRequestBase` schema to allow `null` folder values in (/admin)/permanent-id-requests listing responses.


psarando:
This is a minor bug fix, so I'm going to merge it now, but reviews still welcome.

PR #233, opened 2021-11-30
CORE-1930: Add initial data-usage-api route

ianmcorvidae:
For now this is the only one, but I wanted to get it pushed up now so it's easier to be integrating in the UI.


slr71 (approved):
LGTM :shipit:

PR #236, opened 2022-02-02
Fixed up the data types for a few of the resource-usage endpoints

johnworth:
Basically, the switch to numerics in the backend included a switch to storing the totals in strings to preserve precision across programming languages and serialization formats. This just changes the validation schema to expect strings instead of integers.


slr71 (approved):
Looks good to me

PR #241, opened 2022-04-15
Cache JWKs for up to a day to avoid roundtrips to Keycloak

ianmcorvidae:
I think we could in theory improve this by decoding the JWT header to pull out a key ID and having the cached certs be keyed by that ID, and then only check against certs with a matching key ID. but it probably largely doesn't matter, especially since I think most of the time there's only one key in our use cases.

### DIFF
--- a/src/terrain/util/keycloak_oidc.clj
+++ b/src/terrain/util/keycloak_oidc.clj
@@ -1,11 +1,11 @@
 (ns terrain.util.keycloak-oidc
   (:use [slingshot.slingshot :only [try+ throw+]])
   (:require [clojure.string :as string]
-            [clojure.tools.logging :as log]
             [clojure-commons.jwt :as jwt]
             [clojure-commons.response :as resp]
             [terrain.clients.keycloak :as kc]
-            [terrain.util.config :as config]))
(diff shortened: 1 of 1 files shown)


src/terrain/util/keycloak_oidc.clj
```diff
@@ -29,14 +29,36 @@
...
+(def cache-ttl (* 24 60 60 1000)) ;; 1 day, milliseconds
+
+(defn- update-cert-cache
+  []
+  (otel/with-span [s ["update-cert-cache"]]
+    (let [new-certs (kc/get-oidc-certs)]
+      (reset! cached-certs-time (System/currentTimeMillis))
+      (reset! cached-certs new-certs))))
+
+(defn- validate-with-cache
```
ianmcorvidae, line 43: in short, here, theoretically:

check if there's a cached value and it's not expired
if there is, try to validate with it
if either thing above failed, validate again with an updated cache, and keep/return the error this time if there is one

slr71 (approved):
Looks good to me! 👍

ianmcorvidae:
Cool, thanks for giving it a look before you had to head out!

PR #242, opened 2022-05-12
Make the reason field optional in the waiting state object

johnworth:
I can't find a good way to test this in QA, but this appears to be the field that is causing issues in production. Apparently there is a way for it to be missing.


psarando (approved):
LGTM :shipit:

PR #244, opened 2022-06-21
Start on adding Keycloak service account support to terrain

ianmcorvidae:
I think this is the basic stuff that's needed. It's not used anywhere yet, and I could use some recommendations if there's some endpoint I should add for testing or such.

Still, I think this is ready enough to be reviewed for the structure of the thing!

### DIFF
--- a/src/terrain/util/keycloak_oidc.clj
+++ b/src/terrain/util/keycloak_oidc.clj
@@ -7,24 +7,44 @@
             [terrain.util.config :as config]
             [otel.otel :as otel]))
 
+(defn- is-service-account?
+  [{:keys [preferred_username]}]
+  (string/starts-with? preferred_username "service-account-"))
+
 (defn user-from-token
   "Extracts user information from a Keycloak OIDC token."
   [{:keys [jwt-claims]}]
-  {:shortUsername (:preferred_username jwt-claims)
-   :username      (str (:preferred_username jwt-claims) "@" (config/uid-domain))
-   :email         (:email jwt-claims)
-   :firstName     (:given_name jwt-claims)
-   :lastName      (:family_name jwt-claims)
-   :commonName    (:name jwt-claims)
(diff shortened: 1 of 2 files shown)


src/terrain/util/keycloak_oidc.clj
```diff
@@ -7,24 +7,44 @@
...
+  (if (is-service-account? jwt-claims)
+    {:username (:preferred_username jwt-claims)
+     :roles (:roles (:realm_access jwt-claims))}
+    nil))
 
 (def ^:private required-claims
   [:preferred_username :email :given_name :family_name :name :entitlement])
 
+(def ^:private required-service-account-claims
+  [:preferred_username :entitlement :realm_access])
```
slr71, line 39: I think that `:entitlement` can be removed from the list of required claims for service accounts. This field is present because of the way that Keycloak is configured, but chances are that it will always be empty for service accounts.

slr71 (approved):
This looks good. I had one suggestion to remove a required claim for service accounts, but that's the only thing that I caught.

This is the endpoint that needs this feature: https://de.cyverse.org/terrain/docs/index.html#!/admin45qms/put_terrain_admin_qms_users_username_plan_plan_name

src/terrain/util/keycloak_oidc.clj
ianmcorvidae: Good to know. I'd been wondering about it and figured I'd keep it since it was there and in the older `required-claims`, but I'll remove it since it's probably useless (and after all, we're not using it).

Thanks!

ianmcorvidae:
Ah, that makes sense. It shouldn't be hard to add a middleware like that, and creating that extra endpoint sounds easy enough as well.

I'll probably do a second PR just to isolate things that you've already reviewed, maybe try to get it up later today since you're going on vacation soon.

PR #246, opened 2022-07-20
CORE-1755: added the endpoint to allow service accounts to update user plans

slr71:


slr71:
I found the problem that I was encountering yesterday. The `/admin` element had been removed from some URL paths in the QMS service, and Terrain hadn't  been updated yet. I updated the QMS client namespace to also remove that element from the service calls.

johnworth (approved):
:shipit

psarando (commented):
LGTM 👍

slr71:
Thanks for the reviews!

PR #247, opened 2022-07-28
Check data overage

johnworth:
Adds middleware to check for a data overage before allowing users to upload and save files through the DE.

### DIFF
--- a/src/terrain/routes/data_usage_api.clj
+++ b/src/terrain/routes/data_usage_api.clj
@@ -21,9 +21,5 @@
          :summary schema/UserCurrentDataSummary
          :description schema/UserCurrentDataDescription
          :return schema/UserCurrentDataTotal
-         (ok (dua/user-current-usage (:username current-user))))
-
-       (GET "/overage" []
-         :middleware [require-authentication])))))
-
+         (ok (dua/user-current-usage (:username current-user))))))))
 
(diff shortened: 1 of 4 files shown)


src/terrain/routes/data_usage_api.clj
```diff
@@ -10,17 +10,20 @@
...
+     (context "/data" []
+       (GET "/current" []
+         :middleware [require-authentication]
+         :summary schema/UserCurrentDataSummary
+         :description schema/UserCurrentDataDescription
+         :return schema/UserCurrentDataTotal
+         (ok (dua/user-current-usage (:username current-user))))
+
+       (GET "/overage" []
+         :middleware [require-authentication])))))
```
ianmcorvidae, line 27: Not sure what's going on here? I'm honestly not sure what compojure will do with an empty route like this.

ianmcorvidae (approved):
One thing that possibly creeped in unintentionally, but the main stuff of the PR (middleware, and adding the middleware to stuff) looks good to me

src/terrain/routes/data_usage_api.clj
johnworth: I decided to remove it. It was originally going to be in there for debugging purposes, but it doesn't seem necessary.

johnworth: Also, thanks for catching that.

PR #252, opened 2022-10-19
CORE-1800: add the last modified timestamps to the quota and data usa…

slr71:
…ge records in the resource usage summary endpoint


slr71:
Thanks for yet another quick review ⚡

PR #255, opened 2022-11-22
CORE-1809 Add app_version label to FullInstantLaunch schema

psarando:
This PR will add the  `app_version` label to the `FullInstantLaunch` response schema, and must be merged after cyverse-de/app-exposer#51.


slr71 (approved):
Looks good to me!

psarando:
Thanks for the review! 👍 

I also updated the schema to make the `app_version` key optional, so that the dashboard aggregator doesn't have to be updated for now.

Additional reviews still welcome 🙇

PR #256, opened 2022-11-29
Remove unused resource-usage-api endpoints

johnworth:
This removes most of the resource-usage-api endpoints that aren't currently used by Sonora. The summary endpoint is left in place since it's still used and is supported by the newer version of resource-usage-api. All data-usage-api endpoints are left in place.


slr71 (approved):
Looks good to me! :shipit:

PR #262, opened 2023-01-10
CORE-1834: standardize error responses from QMS endpoints

slr71:
This change standardizes the error response bodies from all QMS endpoints in Terrain so that they more closely resemble error response bodies from other endpoints in Terrain, for example:

Before:
```
$ curl -sH "$AUTH_HEADER" "https://qa.cyverse.org/terrain/admin/qms/users/sarahr/plan/does.not.exist/quota" -H 'Content-Type: application/json' -d '{"quota":10000}' | jq
{
  "error": "resource type 'does.not.exist' not found",
  "status": "Bad Request"
}
```

After:
```
$ curl -sH "$AUTH_HEADER" "http://localhost:8000/terrain/admin/qms/users/sarahr/plan/does.not.exist/quota" -H 'Content-Type: application/json' -d '{"quota":10000}' | jq
{
  "error_code": "ERR_BAD_REQUEST",
  "reason": "resource type 'does.not.exist' not found"
}
```

### DIFF
--- a/src/terrain/clients/coge.clj
+++ b/src/terrain/clients/coge.clj
@@ -1,6 +1,7 @@
 (ns terrain.clients.coge
   (:use [clojure-commons.core :only [remove-nil-values]]
         [terrain.auth.user-attributes :only [current-user]]
+        [terrain.clients.util :only [with-trap]]
         [slingshot.slingshot :only [throw+ try+]])
   (:require [cemerick.url :as curl]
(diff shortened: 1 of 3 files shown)


tedgin (approved):
They look good to me.

psarando (approved):
LGTM 👍

slr71:
Thanks for the reviews!

PR #265, opened 2023-01-27
CORE-1841, CORE-1842: Add endpoints for listing and adding add-ons

johnworth:
This PR adds endpoints for listing and adding add-ons will be applicable to subscriptions.

As part of this PR, support for posting requests to NATS was added along with support for using the Java classes generated from the protocol buffer definitions in the https://github.com/cyverse-de/p/ repository.

### DIFF
--- a/k8s/terrain.yml
+++ b/k8s/terrain.yml
@@ -48,6 +48,18 @@ spec:
             items:
               - key: terrain.properties
                 path: terrain.properties
+        - name: nats-client-tls-pkcs8
+          secret:
+            secretName: nats-client-tls-pkcs8
+            items:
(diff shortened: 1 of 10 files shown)


slr71 (approved):
Looks good to me! :shipit: I would like to see the ability to connect to NATS without SSL or authentication, just to make local development testing easier. That can happen later, though.

psarando (approved):
I didn't catch any issues.

Though I would also like the ability to connect to NATS without SSL or authentication, to make local development testing easier 👍

johnworth:
The ability to turn off TLS should be in there now.

PR #83, opened 2019-02-15
CORE-5: Swagger docs for terrain.routes.comments

slr71:

PR #268, opened 2023-02-04
Group data sharing

ianmcorvidae:
Depends on cyverse-de/common-swagger-api#83

Doesn't currently include analysis sharing. I'll do that next.

### DIFF
--- a/src/terrain/services/sharing.clj
+++ b/src/terrain/services/sharing.clj
@@ -172,12 +172,25 @@
       path-list
       (str "unable to send unshare error notification for " unsharee))))
 
+(defn- get-user-from-subject
+  [subject]
+  (condp = (:source_id subject)
+    "ldap"  (:id subject)
+    "g:gsa" (str "@grouper-" (:id subject))
+    nil))
+
+(defn- translate-user-for-irods
+  [share-unshare]
+  (let [provided-user (:user share-unshare)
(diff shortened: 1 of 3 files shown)


src/terrain/services/sharing.clj
```diff
@@ -172,12 +172,25 @@
...
+        provided-subject (:subject share-unshare)]
+    (or provided-user (get-user-from-subject provided-subject))))
+
 (defn- share-with-user
   "Forwards share requests to data-info from the user and list of paths and permissions in the given
    share map, sending any success notifications to the users involved, and any error notifications
    to the current user."
   [share]
-  (let [user (:user share)
+  (let [user (translate-user-for-irods share)
```
ianmcorvidae, line 193: I figure that if for some reason something passes neither of user/subject, this'll just spit out a nil and that'll throw an error when it tries to pass it down the chain. I could also be more explicit about validation if desired though.

slr71 (approved):
Looks good to me! 👍

ianmcorvidae:
Thanks for the review! I think with a release happening tomorrow, I'll hold off on merging this for right now. We don't have the UI stuff in place anyway, so the only user-facing effect it could have would be breaking stuff if I did something wrong.

psarando (approved):
LGTM 👍

PR #270, opened 2023-02-20
CORE-1878: Add service account accessible subscription add-on endpoints

johnworth:
The cyverse-subscription-updater service account needs to be able to do the following:
* Get a user’s subscription ID from the plan endpoint.
* List available add-ons.
* Add an add-on to a subscription.
* Delete an add-on from a subscription.
* Update an add-on applied to a subscription.
* List add-ons applied to a subscription.
* Get information about a single add-on applied to a subscription.

This should not change the other endpoint(s) already available to the same service account.

JIRA: https://cyverse.atlassian.net/browse/CORE-1878

### DIFF
--- a/src/terrain/routes/qms.clj
+++ b/src/terrain/routes/qms.clj
@@ -215,11 +215,71 @@
      (context "/users" []
        (context "/:username" []
          :path-params [username :- schema/Username]
-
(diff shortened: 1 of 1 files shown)


slr71:
Looks good to me! :shipit:

PR #272, opened 2023-04-10
CORE-1893 Update clj-jargon to v3.1.0

psarando:
This PR will also update calls to `clj-jargon.item-info/trash-base-dir`, which no longer requires a user param, so that items can be put into the same trash paths as used by `irm`.


ianmcorvidae (approved):
Looks good. Boy would I like to remove clj-jargon from terrain at some point, but given how long I've been saying that...

psarando:
Thanks for the review! 👍

PR #275, opened 2023-08-23
CORE-1951: added middleware to ensure that a workspace is created for…

slr71:
… all authenticated users

The result of the HTTP call is memoized in order to eliminate unnecessary HTTP requests. Once we know that the workspace exists, there's no need to make another request to ensure that it still exists.

### DIFF
--- a/src/terrain/clients/apps/raw.clj
+++ b/src/terrain/clients/apps/raw.clj
@@ -1393,13 +1393,19 @@
                 :body         body
                 :content-type :json})))
 
+(def workspace-for-user
+  (memoize (fn [params]
+             (-> (apps-url "workspaces")
+                 (client/get (disable-redirects {:query-params params
+                                                 :as           :json}))
+                 :body))))
+
 (defn bootstrap
   []
-  (:body
-   (client/get (apps-url "bootstrap")
-               (disable-redirects
-                {:query-params (secured-params)
-                 :as           :json}))))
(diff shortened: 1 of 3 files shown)


src/terrain/clients/apps/raw.clj
```diff
@@ -1393,13 +1393,19 @@
...
   []
-  (:body
-   (client/get (apps-url "bootstrap")
-               (disable-redirects
-                {:query-params (secured-params)
-                 :as           :json}))))
+  (-> (apps-url "bootstrap")
+      (client/get (disable-redirects {:query-params (secured-params)
+                                      :as           :json}))
+      :body))
```
ianmcorvidae, line 1408: Just confirming this in case I missed something, but this looks like pure rearrangement? It does look better the changed way, just leaving a note in case there's something more I should look at here.

ianmcorvidae (approved):
The changes look good to me. Presumably with the memoization it shouldn't cause much of a performance hit to do this on every/most requests.

src/terrain/clients/apps/raw.clj
slr71: Correct. I was originally going to cache responses from the Bootstrap endpoint, but it turns out that web hooks are included in the response body. I decided against doing that, but I'd already refactored the function at that point. I thought I'd just leave the changes in place. 😁

slr71:
Thanks for the review! 👋

PR #276, opened 2023-11-14
Core 1962

sboleyn:
Changes to add a feature flag for search in Terrain. 

I'm not sure if this implementation meets the expected outcome for the brief regarding the depth of details provided to the user/ is being implemented when logically needed. Any feedback would be great.


slr71:
This looks good to me! 👍 :shipit:

PR #84, opened 2019-02-19
CORE-6: added Swagger docs to the endpoints in `terrain.routes.data`

slr71:

PR #277, opened 2024-01-03
Pass through parameters for tool details requests so include-defaults comes through

ianmcorvidae:
(depends on cyverse-de/common-swagger-api#84 and this branch will be updated to non-snapshot before merge)


slr71 (approved):
Looks good to me! :shipit:

PR #278, opened 2024-02-23
CORE-1983: add support for multiyear subscriptions

slr71:
This change introduces a couple of new concepts: Subscription Periods, and consumable resource types.

A subscription period refers to the period of time during which the subscription is active. All of our subscriptions are active for one year by default, so a single subscription period is equivalent to one year. With this change, a user may purchase a subscription for multiple subscription periods. For example, if a user chooses to purchase a three year subscription, the user would be given access to the allocations associated with their subscription for three years.

A consumable resource type is a type of resource is not reclaimable. For example, CPU hours can't be reclaimed because once a unit of compute time has been used, it can never be used again. Conversely, storage space can be reclaimed because deleting a file makes the space used to store the file available to be used again. Whether a resource type can be consumed or not is important when creating a new multiyear subscription. In general, we will want to multiply the default allocation associated with a consumable resource, but not a non-consumable resource, by the number of subscription periods when we're computing the quotas for a multiyear subscription.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -66,7 +66,8 @@
             [jonase/eastwood "0.3.10"]
             [reifyhealth/lein-git-down "0.4.1"]]
   :profiles {:dev     {:dependencies [[clj-http-fake "1.0.3"]]
-                       :resource-paths ["conf/test" "test-resources"]}
+                       :resource-paths ["conf/test" "test-resources"]
+                       :jvm-opts ["-Dotel.javaagent.enabled=false"]}
              :uberjar {:aot :all}}
   :main ^:skip-aot terrain.core
   :ring {:handler terrain.core/dev-handler
@@ -79,4 +80,6 @@
                  ["sonatype-releases"
                   {:url "https://oss.sonatype.org/content/repositories/releases/"}]
                  ["public-github" {:url "git://github.com" :protocol :https}]]
-  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/terrain-logging.xml" "-javaagent:./opentelemetry-javaagent.jar" "-Dotel.resource.attributes=service.name=terrain"])
+  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/terrain-logging.xml"
+             "-javaagent:./opentelemetry-javaagent.jar"
+             "-Dotel.resource.attributes=service.name=terrain"])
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -66,7 +66,8 @@
             [jonase/eastwood "0.3.10"]
             [reifyhealth/lein-git-down "0.4.1"]]
   :profiles {:dev     {:dependencies [[clj-http-fake "1.0.3"]]
-                       :resource-paths ["conf/test" "test-resources"]}
+                       :resource-paths ["conf/test" "test-resources"]
+                       :jvm-opts ["-Dotel.javaagent.enabled=false"]}
```
slr71, line 70: I was a little afraid to try to disable the Java agent in the general case even though I was relatively certain that it wouldn't affect runtime behavior. Putting this option in the development profile eliminates most of the errors that we've been seeing, though.

project.clj
```diff
@@ -79,4 +80,6 @@
                  ["sonatype-releases"
                   {:url "https://oss.sonatype.org/content/repositories/releases/"}]
                  ["public-github" {:url "git://github.com" :protocol :https}]]
-  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/terrain-logging.xml" "-javaagent:./opentelemetry-javaagent.jar" "-Dotel.resource.attributes=service.name=terrain"])
+  :jvm-opts ["-Dlogback.configurationFile=/etc/iplant/de/logging/terrain-logging.xml"
+             "-javaagent:./opentelemetry-javaagent.jar"
+             "-Dotel.resource.attributes=service.name=terrain"])
```
slr71, line 85: I just broke up this line so that it would fit the width of my editor.

src/terrain/routes/schemas/qms.clj
```diff
@@ -169,8 +172,11 @@
   {(optional-key :force)
    (describe String "True if the subscription should be created even if the user already has a higher level plan")})
 
+;; The :periods parameter uses s/Int so that we can get automatic coercion.
 (defschema AddSubscriptionParams
-  {:paid (describe Boolean "True if the user paid for the subscription")})
+  {:paid                    (describe Boolean "True if the user paid for the subscription")
+   (optional-key :periods)  (describe s/Int "The number of subscription periods")
```
slr71, line 178: Coercion wasn't working when I tried to use `java.lang.Integer`. Part of me thinks that we should default to using `s/Int` for schemas, although I didn't make that change.

psarando (approved):
LGTM :shipit: 👍

slr71:
Thanks for the review!

PR #281, opened 2024-03-12
CORE-1967 Allow anonymous access in /filesystem/metadata endpoints

psarando:
This PR will add the `secured-filesystem-metadata-routes` to the `optionally-authenticated-routes`, allowing them to be accessed without the `/secured` prefix, and allowing `anonymous` access for the GET endpoints.

POST and CSV download endpoints still require authentication.

This is required in order for Local Contexts Labels and Notices to be viewed by logged out users for public data under Community Data folders.

### DIFF
--- a/src/terrain/util/transformers.clj
+++ b/src/terrain/util/transformers.clj
@@ -40,8 +40,10 @@
   "Generates a set of query parameters to pass to a remote service that requires
    the username of the authenticated user."
   ([]
-     (user-params {}))
+   (user-params {}))
   ([existing-params]
-     (assoc existing-params :user (:shortUsername current-user)))
+   (as-> (add-current-user-to-map {}) m
+         (select-keys m [:user])
+         (merge existing-params m)))
   ([existing-params param-keys]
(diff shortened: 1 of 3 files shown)


src/terrain/util/transformers.clj
```diff
@@ -40,8 +40,10 @@
   "Generates a set of query parameters to pass to a remote service that requires
    the username of the authenticated user."
   ([]
-     (user-params {}))
+   (user-params {}))
   ([existing-params]
-     (assoc existing-params :user (:shortUsername current-user)))
+   (as-> (add-current-user-to-map {}) m
+         (select-keys m [:user])
+         (merge existing-params m)))
```
psarando, line 47: The `add-current-user-to-map` function will substitute `anonymous` for the `:user` field if `:shortUsername` is empty in `current-user` (updated in #177).

slr71 (approved):
Looks good to me! 👍

psarando:
Thanks for the review! 👍

PR #282, opened 2024-03-12
DD-18: added a direct dependency on the slf4j API

slr71:
A newer version of `slf4j-api` was being included by a transitive dependency, which broke logging. I added the explicit dependency to fix logging until we can officially migrate to the newer version.


slr71:
I'm going to merge this immediately. Reviews are still welcome, though.

PR #283, opened 2024-03-25
CORE-1993: fix schema validation errors in subscription add-on endpoints.

slr71:


slr71:
I need to create an add-on, so I'm going to merge this immediately. Reviews are still welcome, though.

PR #285, opened 2024-07-30
Trim leading @ off the UID domain, just in case

ianmcorvidae:
Unsure it's that likely we'll pass in @ signs, but may as well remove them if we do, since the other code assumes there won't be one.


slr71 (approved):
LGTM

PR #286, opened 2024-08-21
CORE-2006: dependency updates.

slr71:
This took a bit longer than I expected. 😅 

The primary changes are as follows:

- Updated most dependencies. There are some incompatibilities between the version of [compojure-api][1] that we use and the latest version of [ring][2]. Upgrading to a version of compojure-api that is compatible requires switching to the latest major version, and there are some fairly big differences between the two major versions. Doing that is going to require more time than I have, so I skipped updating compojure-api and ring. This also requires me to avoid updating [ring-http-response][3] because it depends on the latest verison of ring.
- Removed OpenTelemetry support.
- Renamed the test `logback.xml` file in order to eliminate some warnings that were showing up when compiling or linting the code.
- Removed several functions in the `terrain.persistence.search` namespace, which also eliminated calls to several functions in [elastisch][4] that had breaking API changes in the version that we migrated to. (Be careful with the README for Elastisch; it contains links to some defunct URLs that now have squatters.)

I did some cursory tests using `skaffold dev`, and everything appeared to be working okay. I didn't test very thoroughly, though. My biggest concerns are possible unintended code changes that might have occurred when I was removing OpenTelemetry support and the possibility that I might have missed some API changes in Elastisch when I migrated to the latest version.

[1]: https://github.com/metosin/compojure-api
[2]: https://github.com/ring-clojure/ring
[3]: https://github.com/metosin/ring-http-response
[4]: https://github.com/clojurewerkz/elastisch


psarando (approved):
LGTM 👍

slr71:
Thanks for the review!

PR #287, opened 2024-09-04
Trim the :error key in preferences when other keys are present

ianmcorvidae:
(also, bring back the proper schemas for preferences, now that we know what was actually going on there, namely that the presence of an `:error` key in stored preferences was changing the schema to the error one, ironically, in error)


psarando (approved):
LGTM :shipit: 👍

ianmcorvidae:
Cool, thanks for giving it a look!

PR #288, opened 2024-09-26
CORE-2013: removed support for CAS and WSO2 authentication.

slr71:
While I was at it, I also fixed a bunch of lint warnings.

Most of the lint warnings were in one of these categories:

- Symbols that couldn't be resolved because they were defined by non-standard macros. For these cases, I either found a way to tell `clj-kondo` to handle the definitions or ended up adding declarations to get rid of the warnings.
- Using `:use` instead of `:require` to include other namespaces. In this case, I simply switched to using `:require` instead.
- Unused variable bindings. In most of these cases, I simply prefixed the variable name with an underscore.
- Type hints on the function instead of the parameter vector. In these cases, I moved the type hints.


psarando (approved):
I didn't catch any issues 👍

slr71:
Thanks for the review! I'm going to merge this now! 👍

PR #290, opened 2024-10-17
CORE-1955 Add `PUT /apps/:system/:id/versions` endpoint passthrough

psarando:


ianmcorvidae (approved):
Looks reasonable to me. I thought that `:form-params` wasn't right and it should be `:body`, but the other functions in that namespace clearly do it the same way, so it must work either way with clj-http and the other args as we have them. :shipit:

slr71 (commented):
LGTM :shipit:

psarando:
Thanks for the reviews! 👍

PR #291, opened 2024-10-18
passthrough endpoints for new instant launch non-admin & updating paths for admin

ianmcorvidae:
pairs with cyverse-de/app-exposer#70


ianmcorvidae:
I think that the sonora changes aren't required for merging this and the app-exposer PR, because the existing code already uses admin endpoints in terrain (this just changes where they point in app-exposer). A later PR for actual user-created instant launches will be sonora-only, I think.

ianmcorvidae:
(review still welcome, of course)

PR #292, opened 2024-11-20
Fetch login info (IP, session ID) from Keycloak for recording in the DE database

ianmcorvidae:
This uses a keycloak admin API client to fetch the user's current session(s) from there and pass the information down to apps for storing in the database.

There is one TODO in the code but I think this can be reviewed without fixing it for now.

### DIFF
--- a/src/terrain/clients/apps/raw.clj
+++ b/src/terrain/clients/apps/raw.clj
@@ -1277,8 +1277,8 @@
                    :content-type :json}))))
 
 (defn record-login
-  [ip-address user-agent]
-  (let [params (remove-nil-values {:ip-address ip-address :user-agent user-agent})]
+  [ip-address session-id login-time]
+  (let [params (remove-nil-values {:ip-address ip-address :session-id session-id :login-time login-time})]
     (:body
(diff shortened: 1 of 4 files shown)


src/terrain/clients/apps/raw.clj
```diff
@@ -1277,8 +1277,8 @@
                    :content-type :json}))))
 
 (defn record-login
-  [ip-address user-agent]
-  (let [params (remove-nil-values {:ip-address ip-address :user-agent user-agent})]
+  [ip-address session-id login-time]
+  (let [params (remove-nil-values {:ip-address ip-address :session-id session-id :login-time login-time})]
```
psarando, line 1281: Since `/users/logout` was removed in cyverse-de/apps#277, the `record-logout` function should also be removed from this namespace.

src/terrain/services/bootstrap.clj
```diff
@@ -60,7 +66,7 @@
   [ip-address user-agent]
```
psarando, line 66: Looks like the `ip-address` param can be removed now.

PR #293, opened 2024-12-04
CORE-2016: support for scheduled quota default and subscription rate changes.

slr71:
The primary reason for this change is to allow us to schedule rate and default quota limit changes for the future so that they automatically become active at a time that we specify. For example, we currently have some rate and default quota limit changes planned for January 1st, but we don't want to have to log in on New Years Day to make the changes. Having scheduled changes in the system means that we can set up the changes in advance and let them become effective later.

### DIFF
--- a/src/terrain/clients/qms/model.clj
+++ b/src/terrain/clients/qms/model.clj
@@ -62,7 +62,7 @@
     (when (contains? m :default_paid)
       (.setDefaultPaid builder (:default_paid m)))
     (when (contains? m :addon_rates)
-      (.addAllAddonRates builder (addon-rates-from-maps (:addon-rates m))))
+      (.addAllAddonRates builder (addon-rates-from-maps (:addon_rates m))))
     (.build builder)))
 
 (defn add-addon-request-from-map
--- a/src/terrain/clients/qms/model.clj
+++ b/src/terrain/clients/qms/model.clj
@@ -1,0 +1,118 @@
+(ns terrain.clients.qms.model
+  (:require [terrain.util.time :as time])
+  (:import [org.cyverse.de.protobufs
+            AddAddonRequest
+            Addon
+            AddonRate
+            AssociateByUuids
+            ByUuid
+            NoParamsRequest
+            ResourceType
+            SubscriptionAddon
+            UpdateAddonRequest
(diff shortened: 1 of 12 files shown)


src/terrain/clients/qms/model.clj
```diff
@@ -0,0 +1,246 @@
+(ns terrain.clients.qms.model
```
slr71, line 1: This was a somewhat painful method for generating protocol buffers objects, but [pronto](https://github.com/AppsFlyer/pronto) hasn't been updated to work with newer protocol buffers APIs, and taking the time to update it seemed overwhelming. This works for the time being, but it's not really sustainable. We'll have to find a better option in the near future.

src/terrain/util/nats.clj
```diff
@@ -46,16 +39,12 @@
 
 (defn- json-encode
   [o]
-  (json/generate-string o {:key-fn encode-key}))
+  (.print (JsonFormat/printer) o))
```
slr71, line 42: Cheshire was throwing an error for the protocol buffers that are being created now. I suspect that the reason for this is that we're now using actual protocol buffers objects rather than the `ProtoMap` objects generated by [pronto](https://github.com/AppsFlyer/pronto).

psarando (approved):
LGTM 👍 

I didn't catch any issues, just maybe some harmless cruft?

src/terrain/clients/qms_nats.clj
```diff
@@ -51,55 +48,59 @@
...
 (defn get-subscription-addon
   [uuid]
   (as-> {:uuid (str uuid)} r
-    (nats/request (cfg/get-subscription-addon-subject) ByUUID r)
+    (nats/request-json (cfg/get-subscription-addon-subject) (model/by-uuid-request-from-map r))
     (return-keys r [:subscription_addon])))
+
+(let [printer (JsonFormat/printer)
+      npr     (model/new-no-params-request)]
+  (.print printer npr))
```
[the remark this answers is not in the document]
slr71: Ah, yes. Emacs makes it really easy to type forms into the file and execute them for debugging, and I must have forgotten to remove this one. 😬 Thanks for catching this! 👍

slr71:
Thanks for the review!

PR #295, opened 2025-01-15
CORE-2016: removed the unnecessary `consumable` flag from addon updates

slr71:
The `consumable` flag shouldn't have been added to the resource type for an addon update.


slr71:
I'm going to merge this immediately. Reviews are still welcome, though.

PR #296, opened 2025-01-21
CORE-2029: removed the authentication requirements from a few endpoin…

slr71:
…ts that probably shouldn't have had them in the first place


ianmcorvidae (approved):
Looks reasonable to me. No need for these to be authenticated that I can think of, certainly.

slr71:
Thanks for the review!

PR #297, opened 2025-01-25
CORE-2016: fixed a bug that was preventing addon rates from being updated.

slr71:
This was simply a code change that should have made but wasn't.


slr71:
I'm going to merge this now. Reviews are still welcome.

PR #298, opened 2025-02-05
Allow unauthenticated file downloads to download as the anonymous user

ianmcorvidae:
I think this is all that's needed, but I can't really test this locally because NATS certificates are a pain. So close reviews are definitely welcome.


slr71 (approved):
This looks complete as far as I can tell. Testing this in QA seems reasonable to me. 👍

ianmcorvidae:
Cool. I'll give that a go then, to save myself messing with NATS at all, heh.

PR #299, opened 2025-02-05
update the Skaffold schema version to v3

slr71:
The primary purpose of this is to allow the platform to be specified for cross-compilation.


slr71:
I'm going to merge this immediately so that I can see if it works with CI/CD. Reviews are still welcome, though.

PR #301, opened 2025-02-12
Use ListLoginsResponse from common-swagger-api rather than one only here

ianmcorvidae:
ref: cyverse-de/common-swagger-api#89

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -38,7 +38,7 @@
                  [org.cyverse/cyverse-groups-client "0.1.9"]
                  [org.cyverse/common-cfg "2.8.3"]
                  [org.cyverse/common-cli "2.8.2"]
-                 [org.cyverse/common-swagger-api "3.4.9-SNAPSHOT"]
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -38,7 +38,7 @@
                  [org.cyverse/cyverse-groups-client "0.1.9"]
                  [org.cyverse/common-cfg "2.8.3"]
                  [org.cyverse/common-cli "2.8.2"]
-                 [org.cyverse/common-swagger-api "3.4.8"]
+                 [org.cyverse/common-swagger-api "3.4.9-SNAPSHOT"]
```
psarando, line 41: Don't forget to bump this to `3.4.9` after it's published.

psarando (approved):
LGTM 👍

project.clj
ianmcorvidae: Of course, yeah. Just easier to get it all up for review first, as you know.

psarando: 💯 Just a friendly (pesky) reminder 😅

PR #305, opened 2025-03-19
CORE-2034 Search metadata by `target-ids`

psarando:
This PR will add a `GET /filesystem/metadata` endpoint, which allows a client to search file and folder metadata AVUs by a set of attributes, values, units, and target-ids.

It also will add a `POST /filesystem/metadata/search` endpoint, which does the same thing, only it accepts search params in the request body instead, calling the new `metadata-client.core/search-avus` method.

This is currently restricted to only searching the DE's metadata database, and not iRODS AVUs.

This PR depends on cyverse-de/metadata-client#7 and cyverse-de/common-swagger-api#90 which should be merged before this PR.


psarando:
I'll go ahead and merge this now, but reviews are still welcome 🙇

PR #310, opened 2025-09-25
CORE-2086: added a service-account endpoint for sending emails.

slr71:
The primary purpose of this PR is to add an endpoint that the Subscription Portal can use to send emails to users on behalf of CyVerse. We wanted this endpoint to be fairly secure, so the `cyverse-emailer` role has to be associated with the account in order for calls to this endpoint to succeed.

## Testing

The test is fairly simple. The first step is to obtain an access token for the service account:

``` bash
$ export AUTH_HEADER="Authorization: Bearer $(curl -su "$CLIENT_ID:$CLIENT_SECRET" https://kc.cyverse.org/auth/realms/CyVerse/protocol/openid-connect/token -d grant_type=client_credentials | jq -r .access_token)"
```

**Note:** I was testing against production because I know that emails from production are working. If I had been testing against the QA environment then the Keycloak realm would have been `CyVerseTest`.

Once that's done, sending the email is fairly easy:

``` bash
$ curl -sH "$AUTH_HEADER" "http://localhost:60000/service/email" -H 'Content-type: application/json' -d '{"to": "sarahr@cyverse.org", "from_addr": "support@cyverse.org", "from_name": "CyVerse Support", "subject": "Testing", "template": "bettertest", "values": {"UserEmail": "sarahr@cyverse.org", "Value": "27"}}' | jq
{
  "status": "OK"
}
```

In this case, the email looks like this:

image

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -7,6 +7,12 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 Terrain is the primary REST API gateway for the CyVerse Discovery Environment. It's a Clojure application that validates
 user authentication and coordinates calls to other web services.
 
+## Code Formatting
+
+- Please follow the [clojure community style guidelines][1] when generating new code.
+- Please try to avoid repeated code when possible.
+- Please try to keep line lenghts to 120 characters or fewer for readability.
+
 ## Common Development Commands
 
 ### Build and Run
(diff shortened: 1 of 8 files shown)


slr71:
@psarando had a suggestion to add a `Bcc` field so that we can be copied on emails sent to users. I'm going to implement that suggestion, so I'm marking this PR as a draft for now.

PR #311, opened 2025-10-02
CORE-2086: add support for `CC` and `BCC` email headers.

slr71:
The purpose of this change is to allow the subscription portal to send courtesy copies of emails to CyVerse so that we have a record of the email.


psarando (approved):
LGTM :shipit: 👍

PR #323, opened 2026-05-06
Add admin APIS for vice-operators

johnworth:
Adds admin API calls for managing operators to terrain. They're all pass-throughs to the app-exposer service. Deployed into QA and tested through the Swagger UI.

### DIFF
--- a/src/terrain/routes/schemas/vice.clj
+++ b/src/terrain/routes/schemas/vice.clj
@@ -2,7 +2,8 @@
   (:require
    [common-swagger-api.schema :refer [describe NonBlankString]]
    [schema.core :refer [Any defschema maybe optional-key]]
-   [schema.core :as s])
+   [schema.core :as s]
+   [schema-tools.core :as st])
   (:import
    [java.util UUID]))
 
@@ -437,10 +438,7 @@
   (describe [OperatorAdminSummary] "Registered operators"))
 
 (defschema UpdateOperatorRequest
-  {(optional-key :name)            (describe NonBlankString "New operator display name")
-   (optional-key :url)             (describe NonBlankString "New HTTP(S) base URL")
-   (optional-key :tls_skip_verify) (describe Boolean "New TLS-skip flag")
-   (optional-key :priority)        (describe Long "New scheduling priority")})
+  (st/optional-keys-schema OperatorConfig))
 
 (defschema OperatorCapacityInfo
   {:maxAnalyses              (describe Long "Maximum concurrent analyses")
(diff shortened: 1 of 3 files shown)


src/terrain/routes/schemas/vice.clj
```diff
@@ -421,3 +421,43 @@
...
+   OperatorConfig))
+
+(defschema OperatorAdminSummaryList
+  (describe [OperatorAdminSummary] "Registered operators"))
+
+(defschema UpdateOperatorRequest
+  {(optional-key :name)            (describe NonBlankString "New operator display name")
+   (optional-key :url)             (describe NonBlankString "New HTTP(S) base URL")
+   (optional-key :tls_skip_verify) (describe Boolean "New TLS-skip flag")
+   (optional-key :priority)        (describe Long "New scheduling priority")})
```
psarando: These all look the same as `OperatorConfig`, so this could defined with `schema-tools.core/optional-keys-schema`:

```suggestion
(defschema UpdateOperatorRequest
  (schema-tools.core/optional-keys-schema OperatorConfig))
```

src/terrain/routes/vice.clj
```diff
@@ -73,6 +73,43 @@
...
+         (PATCH "/" []
+           :body [body vice-schema/UpdateOperatorRequest]
+           :return vice-schema/OperatorAdminSummary
+           :summary "Update an operator"
+           :description "Partial update of an operator identified by UUID. Only fields supplied in the body are changed."
+           (ok (vice/admin-update-operator id body)))
+
+         (DELETE "/" []
+           :summary "Delete an operator"
+           :description "Removes the operator with the given UUID. Idempotent. Fails if jobs still reference the operator."
```
psarando, line 110:
```suggestion
           :description "Removes the operator with the given UUID. Idempotent. Fails if jobs still references the operator."
```

psarando (approved):
I didn't catch any issues 👍 

I just had a couple of minor suggestions.

src/terrain/routes/vice.clj
johnworth: I think in this case jobs is actually plural, so "reference" is correct. If we were dealing with the jobs table itself it would need the change.

src/terrain/routes/schemas/vice.clj
johnworth: Good catch, thanks. Changes should be made.

src/terrain/routes/vice.clj
psarando: Oops, not sure how I just glazed over that `s` on the end of jobs 🤦

PR #324, opened 2026-06-04
Add support for new operator fields

johnworth:
Adds support for the following fields:
* accepting_launches - whether the operator accepts new app launches. Turning this to false allows us to phase out an operator without just terminating a bunch of running analyses.

* deactivated - essentially a soft-delete of an operator.

* supportedGPUModels - the operators can now return a listing of their supportedGPUModels in their capacity endpoint. It's useful to have access to that info through the API.

I noticed the inconsistency in the casing across fields, but upon exploration of the code a bit it's become clear that this is a very common issue in our codebase. Not something I can easily address here, unfortunately.

### DIFF
--- a/src/terrain/routes/schemas/vice.clj
+++ b/src/terrain/routes/schemas/vice.clj
@@ -434,13 +434,21 @@
 (defschema OperatorAdminSummary
   (merge
    {:id (describe UUID "Server-assigned UUID of the operator")}
-   OperatorConfig))
+   OperatorConfig
+   {:accepting_launches (describe Boolean "Whether the operator accepts new launches; false drains it but leaves it in service")
+    :deactivated        (describe Boolean "Whether the operator is fully removed from the live pool; takes precedence over accepting_launches")}))
 
 (defschema OperatorAdminSummaryList
   (describe [OperatorAdminSummary] "Registered operators"))
 
 (defschema UpdateOperatorRequest
-  (st/optional-keys-schema OperatorConfig))
+  (merge
+   (st/optional-keys-schema OperatorConfig)
+   {(optional-key :accepting_launches)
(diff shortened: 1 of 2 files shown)


src/terrain/routes/schemas/vice.clj
```diff
@@ -434,13 +434,21 @@
...
 
 (defschema UpdateOperatorRequest
-  (st/optional-keys-schema OperatorConfig))
+  (merge
+   (st/optional-keys-schema OperatorConfig)
+   {(optional-key :accepting_launches)
+    (describe Boolean "Whether the operator accepts new launches; false drains it but leaves it in service")
+
+    (optional-key :deactivated)
+    (describe Boolean "Whether the operator is fully removed from the live pool; takes precedence over accepting_launches")}))
```
psarando, line 451: These new keys could be added to `OperatorConfig`, then they wouldn't have to be repeated here and in `OperatorAdminSummary`.

src/terrain/routes/schemas/vice.clj
```diff
@@ -434,13 +434,21 @@
 (defschema OperatorAdminSummary
   (merge
    {:id (describe UUID "Server-assigned UUID of the operator")}
-   OperatorConfig))
+   OperatorConfig
+   {:accepting_launches (describe Boolean "Whether the operator accepts new launches; false drains it but leaves it in service")
+    :deactivated        (describe Boolean "Whether the operator is fully removed from the live pool; takes precedence over accepting_launches")}))
```
psarando, line 439: I should have commented here: I meant the non-optional version of these new keys could be added to `OperatorConfig`...

psarando (approved):
LGTM 👍 

I just caught a minor DRY issue.

PR #325, opened 2026-06-04
Allow the ingresses key through the VICE resources listing

johnworth:
## Summary

The `GET /api/admin/vice/resources` and non-admin `GET /api/vice/resources` endpoints return `FullResourceListing` as their `:return` schema. Because plumatic/schema strips keys not declared in the schema, the `:ingresses` key produced by app-exposer's `ResourceInfo` was being dropped from the response.

## Changes

- Add an `Ingress` schema mirroring app-exposer's `reporting.IngressInfo` (`MetaInfo` fields via `BaseListing`, plus `defaultBackend` and `rules`).
- Wire `:ingresses` into `FullResourceListing`, ordered between `:services` and `:routes` to match app-exposer's `ResourceInfo`.
- `IngressRule` documents `:host` as an optional key and adds an `s/Keyword s/Any` catch-all, so any additional Kubernetes `IngressRule` fields pass through untouched.

### DIFF
--- a/src/terrain/routes/schemas/vice.clj
+++ b/src/terrain/routes/schemas/vice.clj
@@ -380,11 +380,22 @@
    BaseListing
    {:rules (describe [HTTPRouteRule] "Rules used to match requests with the route")}))
 
+(defschema IngressRule
+  {(optional-key :host) (describe String "The fully qualified domain name the rule applies to")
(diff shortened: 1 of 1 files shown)


psarando (approved):
LGTM 👍

PR #326, opened 2026-06-08
CORE-2158: Add OIDC Authorization Code Flow support to terrain

slr71:
Terrain has only ever consumed Keycloak OIDC bearer tokens. This adds support for acting as an OIDC relying party using the browser-based Authorization Code Flow, and integrates it into the Swagger UI.

New unsecured endpoints under /oidc:
  - GET  /oidc/login    redirect to Keycloak's authorize endpoint
  - GET  /oidc/callback verify state, exchange code, set token cookies
  - POST /oidc/refresh  refresh-token grant, reset cookies
  - GET  /oidc/logout   clear the token cookies

After the callback exchanges the code for tokens, the access and refresh tokens are stored in HttpOnly cookies and the browser is redirected to a landing URI. The auth middleware now also accepts the access token from the cookie, so requests carrying it authenticate like a bearer token.

The OIDC redirect and landing URIs are derived from a single new public terrain.base-uri setting so the two can't drift and there's less to misconfigure.

For Swagger, ring-swagger-ui is upgraded to 5.x and an oauth2 securityDefinition (accessCode flow) is added pointing at Keycloak. A config-driven swagger-initializer.js calls initOAuth so the Authorize button can perform the flow.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -29,6 +29,7 @@
                  [org.cyverse/common-cfg "2.8.3"]
                  [org.cyverse/common-cli "2.8.2"]
                  [org.cyverse/common-swagger-api "3.4.17"]
+                 [metosin/ring-swagger-ui "5.20.0"]
                  [org.cyverse/kameleon "3.0.10"
                   :exclusion [com.impossibl.pgjdbc-ng/pgjdbc-ng]]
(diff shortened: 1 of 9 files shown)


psarando (approved):
I didn't catch any issues 👍

PR #328, opened 2026-06-11
Add user-level VICE and binary file-overwrite routes for formation

johnworth:
Formation is being retargeted to call terrain (with the caller's bearer token) instead of hitting app-exposer admin endpoints and iRODS directly. That needs user-level equivalents of a few operations that were previously admin-only or string-only:

- `GET /vice/async-data`, `GET /vice/analyses/{id}/external-id`, and `POST /vice/analyses/{id}/exit`, backed by a new `terrain.services.vice` namespace. Exit requires the same write permission the apps service enforces for analysis stop; the read-level routes require read permission. All checks are a single permissions-service lookup with `lookup=true`, so grants held via group/team membership are honored the same way apps honors them. The checks live in terrain because app-exposer's user-level handlers do no ownership enforcement.
- `POST /fileio/overwrite`: binary-safe multipart overwrite of an existing file, modeled on wrap-file-upload and wired to the already-existing data-info overwrite client chain (PUT /data/{data-id}). The existing fileio/save endpoint remains UTF-8-string-only.

Also fixes two pre-existing bugs found along the way:

- admin-external-id now reshapes app-exposer's external_id key to the published schema's externalID; the admin route's response validation has been broken since app-exposer fixed its malformed struct tag.
- resolve-test-user passed :user-attribues (typo) to user-from-attributes, leaving with-user test fixtures with a nil shortUsername.

Code review fixes:

- Replaced the original apps-listing + permission-lister authorization in `terrain.services.vice` with the permissions-service lookup described above. The original check missed write grants held via group membership (a team member who could stop an analysis through apps got a 403 from exit), tied authorization to apps listing semantics (analyses excluded from the default listing read as missing even to their owner), and cost up to three HTTP round trips per request — one of them apps' heaviest listing path. The new check is one permissions-service call.
- The async-data not-found error is keyed to the external ID instead of the analysis UUID, so an unauthorized caller can no longer harvest internal analysis IDs from the error message.
- `POST /fileio/upload` now runs the data-overage check before the multipart middleware. Previously the multipart store uploaded the file to the Data Store and *then* the overage check rejected the request, so over-quota users still landed files.
- The shared multipart middleware trims `dest`, matching save/saveas/urlupload.
- Cleanups: the multipart middlewares are plain defns (visible arity, REPL-rebindable store fns), the JSON-stub test helper moved to `terrain.test-fixtures` as the canonical copy, and the exit permission tests are table-driven across all four permission levels plus the no-access case.

### DIFF
--- a/src/terrain/auth/user_attributes.clj
+++ b/src/terrain/auth/user_attributes.clj
@@ -240,7 +240,7 @@
   "Attempts to resolve a test user from either a username or a map of user attributes."
   [user]
   (cond (string? user) (lookup-user user)
-        (map? user)    (user-from-attributes {:user-attribues user})
+        (map? user)    (user-from-attributes {:user-attributes user})
         :else          (cxu/internal-system-error "user must be a string or a map of attributes")))
 
 (defmacro with-user
--- a/src/terrain/clients/app_exposer.clj
+++ b/src/terrain/clients/app_exposer.clj
@@ -111,9 +111,11 @@
   (:body (client/get (app-exposer-url ["vice" "async-data"] query) {:as :json})))
 
 (defn admin-external-id
-  "Gets the external-id for the provided analysis-id. VICE analyses only have a single external-id"
+  "Gets the external-id for the provided analysis-id. VICE analyses only have a single external-id.
+   app-exposer responds with an external_id key; the published schema uses externalID."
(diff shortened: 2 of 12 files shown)


slr71 (approved):
This looks good to me! 👍

PR #329, opened 2026-07-01
Fix upload of files with non-ASCII names (RFC6532 multipart mode)

johnworth:
## Problem

Uploading a file whose name contains non-ASCII characters (e.g. `test-ä-ö-ü.txt`) stores it with those characters replaced by `?`. Reproduced via the web UI: uploading `test-ä-ö-ü.txt` lands a file named `test-?-?-?.txt`.

## Root cause

The corruption happens in terrain, not sonora, data-info, or iRODS:

- **sonora** pipes the raw multipart request body straight through (`req.pipe(...).pipe(res)`), so the browser's UTF-8 filename reaches terrain intact.
- **terrain** re-parses the upload and re-emits a *new* multipart request to data-info via clj-http in `upload-file` (`src/terrain/clients/data_info/raw.clj`). clj-http defaults to `HttpMultipartMode/STRICT`, which encodes the multipart `filename` parameter as **US-ASCII** and replaces any character it cannot represent with `?` — before data-info ever receives the file.
- data-info / clj-jargon / iRODS were exonerated (Jargon's default protocol encoding is UTF-8; JDK 22/25 default charset is UTF-8). The `?` (rather than mojibake) is the signature of an ASCII-replacement step, which only STRICT-mode multipart encoding performs on this path.

## Fix

Set `:multipart-mode HttpMultipartMode/RFC6532` on the `upload-file` request. RFC6532 encodes the filename as UTF-8, preserving the original characters end-to-end.

`overwrite-file` is intentionally left unchanged: it identifies the target file by UUID in the URL and sends no user-supplied filename (data-info's overwrite handler ignores the multipart filename), so it cannot exhibit this bug.

## Tests

`test/terrain/clients/data_info/raw_test.clj`:

- `multipart-mode-filename-encoding` — table-driven behavioral test that serializes the actual multipart body and asserts RFC6532 preserves `test-ä-ö-ü.txt` while the old STRICT default mangles it (documents the bug and the rationale).
- `upload-file-uses-utf8-multipart-mode` — asserts `upload-file` requests RFC6532 mode and passes the original filename through unchanged (guards the call site against regression).

Verified the wiring test fails if the mode is reverted to STRICT. `lein uberjar` builds cleanly.

## Scope note (download path)

An earlier draft of this PR flagged a download-side defect: data-info builds `Content-Disposition` with the raw filename and no RFC 6266 `filename*` encoding (`data_info/services/entry.clj`). On closer inspection this does **not** affect the reported characters — ä/ö/ü (and the whole Latin-1 supplement) are ≤ U+00FF, and the HTTP header transport charset (ISO-8859-1) coincides with RFC 6266's `filename=` fallback charset, so the byte round-trips correctly and downloads work (verified manually). The only theoretical residual is filenames containing characters above U+00FF, which is unverified, cosmetic at worst, and out of scope here.

### DIFF
--- a/src/terrain/clients/data_info/raw.clj
+++ b/src/terrain/clients/data_info/raw.clj
@@ -6,7 +6,8 @@
             [cheshire.core :as json]
             [clj-http.client :as http]
             [terrain.util.config :as cfg])
-  (:import [clojure.lang IPersistentMap ISeq Keyword]))
+  (:import [clojure.lang IPersistentMap ISeq Keyword]
+           [org.apache.http.entity.mime HttpMultipartMode]))
 
 ;; HELPER FUNCTIONS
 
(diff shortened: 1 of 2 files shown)


psarando (approved):
LGTM 👍

slr71 (approved):
Looks good to me too. 👍

PR #330, opened 2026-07-02
Switch QMS add-on operations from NATS+protobuf to the subscriptions HTTP API

johnworth:
Part of the DE-wide protocol buffers removal. **Deploy after cyverse-de/subscriptions#25** so both transports speak the same JSON dialect.

- The add-on and subscription-add-on calls now use the subscriptions service's existing HTTP endpoints instead of NATS requests built from protobuf classes. New config: `terrain.subscriptions.base-uri` (default `http://subscriptions`) — add it to the deployed terrain.properties if the default doesn't match the environment.
- Request bodies keep the same wire keys the NATS messages used (including the camelCase `updateName`/`updateDescription`/… flags on add-on updates), and responses carry the same embedded `error` object, so route behavior and schemas are unchanged.
- QMS was terrain's only NATS use, so the NATS connection setup, TLS config, and the `jnats`, `cyverse-de-protobufs`, `pronto`, and `less-awful-ssl` dependencies are removed entirely. The protobuf-builder tests are replaced by table-driven tests of the new request shaping.
- The `org.cyverse/cyverse-de-protobufs` Clojars artifact has no remaining consumers after this; p no longer publishes it.


build.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
```
slr71, line 1: I thought that the `Justfile` in this repo managed everything that this file manages. What's the reason for having both?

slr71 (approved):
I had one question about a file that might be innocuous but redundant (unless I've misunderstood something). Everything else looks good to me! 👍

PR #331, opened 2026-07-06
Give the file-upload endpoint a longer, configurable Jetty idle timeout

slr71:
## Problem

Large file uploads over slow/intermittently-stalled connections were cut off with an **I/O timeout after ~200s of socket inactivity**. Terrain's `run-jetty` used Jetty's default 200000 ms idle timeout, which applies to the (long-lived, streaming) upload connection just like any quick API call.

## Change

Rather than raise the global idle timeout (which would also let dead connections on ordinary API endpoints linger), install a Jetty `HttpConfiguration` customizer that:

- keeps a short default idle timeout (`terrain.app.jetty-max-idle-time`, 200s) for all endpoints, and
- raises it (`terrain.app.jetty-upload-idle-time`, 1h) only for the `/secured/fileio/upload` endpoint.

The customizer sets the value per request, so it stays correct across keep-alive connection reuse. Both timeouts are configurable.

Files: `util/jetty.clj` (new), `core.clj` (wire the `:configurator`), `util/config.clj` (new properties).

## Testing

Verified against the QA environment:

- 210s stall during an upload → **HTTP 200** (previously `Idle timeout expired: 200000/200000 ms`).
- 31-minute steady 512 MB upload → **HTTP 200**, full file intact.
- Ordinary authenticated API calls unaffected (they keep the short default timeout).

Pairs with cyverse-de/data-info#79, which does the same for data-info (and fixes the related spurious `ERR_EXISTS` on interrupted uploads).

### DIFF
--- a/src/terrain/util/jetty.clj
+++ b/src/terrain/util/jetty.clj
@@ -5,18 +5,33 @@
    grace period for slow uploads would otherwise apply to every endpoint. Installing an
    HttpConfiguration customizer lets us keep a short default idle timeout while raising it
    only for the (long-lived, streaming) upload endpoint."
-  (:import [org.eclipse.jetty.server Server Connector HttpConnectionFactory HttpConfiguration$Customizer]))
+  (:import [org.eclipse.jetty.server Server Connector HttpConnectionFactory HttpConfiguration$Customizer Request]
+           [java.util.function Consumer]))
 
 (defn- idle-timeout-customizer
   "A per-request Jetty customizer. Sets the connection's idle timeout to upload-idle-ms for
    requests whose path satisfies upload-path?, and to default-idle-ms otherwise (so the value
-   is correct even when a connection is reused across requests)."
+   is correct even when a connection is reused across requests).
+
+   For upload requests it also registers a completion listener that resets the idle timeout
+   back to default-idle-ms once the request finishes. Otherwise the raised timeout would
+   persist on a keep-alive connection until the next request ran customize, letting a
+   post-upload connection linger for upload-idle-ms if the client stalls without sending
+   another request."
   [upload-path? ^long default-idle-ms ^long upload-idle-ms]
   (reify HttpConfiguration$Customizer
     (customize [_ request _response-headers]
-      (let [endpoint (.. request getConnectionMetaData getConnection getEndPoint)
+      (let [^Request request request
+            endpoint (.. request getConnectionMetaData getConnection getEndPoint)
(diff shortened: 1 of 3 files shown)


ianmcorvidae (approved):
LGTM. Poked around a bit because I was thinking there was another endpoint that would create new files, but I can't find it now so I was probably just imagining it

src/terrain/util/jetty.clj
```diff
@@ -0,0 +1,31 @@
...
+(defn- idle-timeout-customizer
+  "A per-request Jetty customizer. Sets the connection's idle timeout to upload-idle-ms for
+   requests whose path satisfies upload-path?, and to default-idle-ms otherwise (so the value
+   is correct even when a connection is reused across requests)."
+  [upload-path? ^long default-idle-ms ^long upload-idle-ms]
+  (reify HttpConfiguration$Customizer
+    (customize [_ request _response-headers]
+      (let [endpoint (.. request getConnectionMetaData getConnection getEndPoint)
+            path     (.. request getHttpURI getPath)]
+        (.setIdleTimeout endpoint (if (upload-path? path) upload-idle-ms default-idle-ms)))
```
johnworth: The raised idle timeout persists on a keep-alive connection after the upload finishes. The customizer mutates the shared `EndPoint` idle timeout and nothing resets it at end-of-request; on an HTTP/1.1 persistent connection the value only returns to `default-idle-ms` when the *next* request runs `customize`.

Scenario: a client POSTs `/secured/fileio/upload`, the upload completes and the response is sent, then the client keeps the connection open without sending another request (or stalls/dies). That socket now sits idle for up to `upload-idle-ms` (1h) instead of the 200s default. Since that same connection is what a browser reuses for subsequent ordinary API calls, this is the "dead connections on ordinary API endpoints linger" behavior the PR description says it avoids — just deferred to post-upload connections, and it can accumulate against fds/selector slots under many concurrent uploads.

If that's acceptable in practice (uploads are infrequent; the connection closes within the hour regardless), a one-line note here would set expectations. Otherwise, reset the endpoint idle timeout back to `default-idle-ms` when the upload request completes.

slr71: Good catch — the raised timeout did persist on the reused connection. Fixed in 527d47f: upload requests now register a `Request` completion listener that resets the endpoint idle timeout back to `default-idle-ms` when the request finishes, so a keep-alive connection returns to the short default rather than lingering for `upload-idle-ms`.

PR #332, opened 2026-07-23
Reach QMS add-ons over HTTP and remove NATS from terrain

johnworth:
## Why

terrain held a NATS connection for exactly one purpose: 9 synchronous request/reply calls to the QMS **add-on** and **subscription-add-on** operations (`clients/qms_nats.clj`). Those same operations are already served over HTTP by the **subscriptions** service, backed by the identical handlers. Switching to HTTP lets terrain drop NATS entirely — no more NATS connection at startup, no NATS TLS/creds/PKCS8 mounts, and one fewer service on the NATS client list.

## What

- **New `terrain.clients.qms-addons`** — an HTTP client mirroring `terrain.clients.qms` (clj-http + `cemerick.url` + `with-trap`). It keeps the **same 9 public fn names, arg lists, and return-key shapes** as the old NATS client, so `routes/qms.clj` only needed its alias repointed (`qms-nats` → `qms-addons`); the ~15 route handlers are unchanged, as are their `require-authentication` / `require-service-account` guards.
- **Request bodies are plain maps matching the subscriptions `p/go/qms` JSON tags.** The two update endpoints use *different* mask casing — camelCase (`updateName`…) for add-ons, snake_case (`update_amount`…) for subscription add-ons — and the client honors each. For the get/delete/update subscription-addon routes, the service reads only the add-on UUID, so the add-on UUID is reused for both path segments (documented inline).
- **Config**: added `config/subscriptions-base-uri` (`terrain.subscriptions.base-uri`, default `http://subscriptions`); removed the whole `terrain.nats.*` config block.
- **Startup**: removed the `nats-connect` bootstrap from `core.clj` — terrain no longer opens a NATS connection at startup.
- **Dead code / deps**: deleted `util/nats.clj`, the protobuf request model `clients/qms/model.clj` + its test, and `util/time.clj` (all only used by the NATS path); dropped the `io.nats/jnats`, `less-awful-ssl`, `com.appsflyer/pronto`, `org.cyverse/cyverse-de-protobufs`, and `clojure.java-time` deps (verified no other users).
- **Tests**: `qms_addons_test.clj` covers per-endpoint URL/method, request-body mask casing (incl. present-but-false vs unset), payload extraction, and error mapping.

## Verification

- `lein check` — exit 0, all namespaces compile without the NATS/protobuf deps.
- `lein test terrain.clients.qms-addons-test` — 6 tests, 28 assertions, 0 failures.
- `clj-kondo` and `lein cljfmt check` — clean.
- Net: **-879 / +259** lines.

## Depends on / rollout

- Pairs with **cyverse-de/subscriptions#26** (valid HTTP error status codes), which should deploy first so terrain gets clean 4xx/5xx.
- The config default (`http://subscriptions`) means this image works before the deployments change lands. The deployments PR then removes terrain's NATS mounts and wires `baseurls_subscriptions`. Do **not** apply the deployments NATS-removal against an old (NATS-based) terrain image.


slr71 (approved):
Looks good to me. 👍

PR #334, opened 2026-08-15
instructions for running Terrain locally

slr71:


johnworth (approved):
LGTM

README.markdown
```diff
@@ -1,9 +1,96 @@
...
+```
+$ kubectl -n $NAMESPACE get secret service-configs -o json | jq -r '.data["terrain.properties"] | @base64d' \
+    > "$CONF"
+```
+
+You should be able to use this file without modification.
+
+#### Forward Connections to Kubernetes
+
+This step requires [kubefwd][5], which auomatically configures port forwarding and host name aliases. In another
```
ianmcorvidae: typo nitpick

```suggestion
This step requires [kubefwd][5], which automatically configures port forwarding and host name aliases. In another
```

README.markdown
```diff
@@ -1,9 +1,96 @@
...
+$ export AUTH_HEADER="Authorization: bearer $(curl -su "$DE_USER" "http://localhost:60000/terrain/token" | jq -r .access_token)"
+```
+
+You can run this command to verify that the command worked:
+
+```
+$ echo $AUTH_HEADER
+```
+
+If the value is `Authorization: berer null` then the authentication didn't work. In most cases, it means that the
```
ianmcorvidae: another typo (unless HTTP auth did something weird again that I don't know about haha)

```suggestion
If the value is `Authorization: bearer null` then the authentication didn't work. In most cases, it means that the
```

ianmcorvidae (approved):
typo nitpicks, looks good otherwise

slr71:
Thanks for the review! 👍

PR #336, opened 2026-08-18
Update dependencies and gate on dependency conflicts

slr71:
Part of a fleet-wide pass over the DE Clojure repos to clear Leiningen's
"Possibly confusing dependencies" reports and bring dependencies current.
**17 warnings → 0.**

## Changes

- `build: normalize plugin versions`
- `deps: update org.cyverse libraries and fix an ignored exclusion`
- `build: fail the build on confusing dependencies`

## The cider-nrepl leak is fixed at the source

Published `org.cyverse` poms used to declare `cider/cider-nrepl` at *compile*
scope, so a development REPL tool was landing on this service's runtime
classpath and inside its uberjar. Those libraries have since been re-released
with it removed, so bumping the pins was enough — **no exclusion workaround
needed**. Verified: 0 cider entries on the resolved classpath and 0 in the
built uberjar.

## A silently-ignored exclusion, now fixed

The `kameleon` dependency used the singular `:exclusion` key:

```clojure
[org.cyverse/kameleon "3.0.10"
 :exclusion [com.impossibl.pgjdbc-ng/pgjdbc-ng]]   ;; not a Leiningen key
```

`:exclusion` is not a Leiningen key, so this has never done anything. The intent
was already satisfied by the direct `pgjdbc-ng` dependency declared on the next
line, so correcting it to `:exclusions` is behaviour-preserving — I verified
`pgjdbc-ng` still resolves to 0.8.9 before and after.

## cljfmt was not actually runnable

This repo declared the deprecated `lein-cljfmt "0.9.2"` coordinates. It now uses
`dev.weavejester/lein-cljfmt` in its own profile — its tree and test2junit's
disagree on which Clojure to use, which trips the gate on a plugin-only
conflict. Format with `lein with-profile +cljfmt cljfmt check`. It currently
reports all source files formatted correctly.

## A note on the jackson versions

`clj-jargon` holds `databind`/`annotations` at the 2.14.1 that `jargon-core
4.3.7.0-RELEASE` brings (pinned `:upgrade false` for iRODS), while
`jackson-core` sits higher because cheshire needs it there. The family is
deliberately not uniform, matching what `main` already resolved. The comment in
`project.clj` warns against "tidying" it — that would break JSON handling at
runtime rather than at build time.

## Verification

```
confusing-dep warnings   17 -> 0
lein deps :tree          exit 0 (default, +dev,+test, +uberjar)
lein test                Ran 19 tests containing 65 assertions., 0 failures
lein uberjar             builds; 0 cider entries in the jar
```


slr71:
I'm going to merge this now in order to get the changes tested as quickly as possible. Reviews are still welcome

PR #337, opened 2026-08-18
Proxy data operations through data-info instead of connecting to iRODS

johnworth:
Depends on cyverse-de/data-info#82 — the endpoints this PR calls have to be deployed first.

terrain depended on `clj-jargon` and `clj-icat-direct` and opened its own connections to both, for a handful of operations data-info otherwise owns. Most of that code was a fork of data-info's that had since drifted — `filesystem/sharing.clj` against data-info's `services/sharing.clj`, `common_paths.clj` against `util/paths.clj`, `garnish/irods.clj` against the `get-types` in `services/stat/jargon.clj`. Two services answering the same question about the same file, able to disagree.

**756 lines deleted, 175 added.**

## Sharing no longer fans out

`services/sharing.clj` called `data/share` once per user/path pair, with single-element lists, and derived per-path success from a `try+ / catch map?` around each call. In-process that was cheap; over HTTP it would have been one round trip per pair.

It now sends one request and reads the per-path outcome data-info reports. The grouping that drives the two notifications is unchanged — `(filter :success ...)` / `(remove :success ...)` still works, it just operates on data-info's items rather than locally synthesized ones.

`data/share` and `data/unshare` keep their throwing contract, because `permanent_id_requests.clj` (5 call sites) and `coge.clj` depend on it and would otherwise silently proceed past a failed share. The new `share-paths`/`unshare-paths` expose per-item outcomes for the one caller that wants them.

## Behavior preserved deliberately

- **`resolve-data-type`** passes `ignore-missing`/`ignore-inaccessible` and falls back to `"file"`. The ICAT lookup it replaces returned `"file"` for a data id that resolved to nothing; without the flags data-info would raise instead.
- **`uuid-accessible?`** likewise, so an inaccessible id reports `false` rather than throwing.
- **Skipped shares still count as successes.** data-info reports them with a `reason`; terrain drops it, which keeps the notifications identical to today.
- **`ensure-dir-created`** checks existence first, since data-info's `create-dirs` raises `ERR_EXISTS` where terrain's `ensure-created` was a no-op.

## What stayed

Files that were entirely iRODS are deleted: `icat.clj`, `uuids.clj`, `sharing.clj`, `create.clj`, `garnish/irods.clj`. Files that were only part iRODS kept their pure halves:

- `stat.clj` → `get-public-data-user` (an authorization policy decision — anonymous fallback for community data) and `do-stat` (already a passthrough)
- `validators.clj` → `not-superuser`
- `common_paths.clj` → the logging macros used in 19 places, plus `id->label` and `user-home-dir`. `user-trash-path` now builds the path directly instead of borrowing clj-jargon's `trash-base-dir`, which was pure string manipulation anyway.

`irods-home`, `irods-user`, `irods-zone` and `fs-community-data` are **not** removed — they're used for path building and identity outside the filesystem package (`directory.clj`, `permanent_id_requests.clj`, `metadata/tags.clj`). The 15 connection and credential properties are gone, and `exception-filters` no longer references the ICAT password since it no longer exists.

Favorites sort columns now map onto data-info's listing vocabulary rather than raw ICAT column names. That translation stays at the gateway, as does subject-to-iRODS-username resolution (`g:gsa` → `@grouper-<id>`), which is Grouper knowledge data-info shouldn't have.

## Testing

`lein test`: 19 tests, 65 assertions, 0 failures. `lein check` clean.

The existing suite doesn't cover these paths, so the real verification is the deploy — the areas to exercise are sharing and unsharing (including the notifications), the favorites listing with each sort column, URL import, and default-output-directory validation.


ianmcorvidae (approved):
This looks pretty reasonable to me, other than any changes that might be necessary based on my data-info PR comments.

slr71 (approved):
Looks good to me! 👍

PR #338, opened 2026-08-21
CORE-2197: add support for metadata filtering to the app search endpo…

slr71:
…ints.

These endpoints are simple pass-through calls to the apps service, so the only change that is required is to bump the common-swagger-api version.


psarando (approved):
:shipit:

psarando (approved):
LGTM 👍

